PR #5, opened 2017-08-24
Work around Pkg.test forcing bounds checks.

maleadt:
Kinda stinks, but does the job, and makes sure even CI runs all tests.


MikeInnes:
Great idea, thanks! It even passes the shiny new CI :+1:

MikeInnes:
I changed this to use a simple `run`. If the tests fail the extra stacktrace isn't a big deal, but it's nice to be able to run Pkg.test locally with output.

PR #16, opened 2017-09-11
add missing useful convert functions

ilkerkesen:
This PR makes the following operations possible,

```julia
using CuArrays
a = rand(Float64, 10, 20)
c1 = convert(CuArray{Float32}, a)
c2 = convert(CuArray{Float32,2}, a)
```


MikeInnes:
Neat, thanks!

PR #19, opened 2017-09-21
move to GPUArrays

SimonDanisch:


SimonDanisch:
Needs https://github.com/JuliaGPU/CUDAnative.jl/pull/119

MikeInnes:
Great. In order to make this a bit more manageable it'd be nice to split it up a bit, i.e. figure out what the minimal patch is to be a `GPUArray` and then add features from there. In particular there are things like FFT functionality that we'll want to do differently (drop the deprecated CUDArt, and probably fold CUFFT itself into this package).

MikeInnes:
For reference, this depends on https://github.com/JuliaGPU/GPUArrays.jl/pull/64.

SimonDanisch:
After moving to the JuliaGPU, this PR lives now in #20

PR #27, opened 2017-10-02
delete ext.jl so the package can be updated

gaika:
To avoid "WARNING: Package CuArrays: skipping update (dirty)..."


MikeInnes:
Good catch, thanks. It would also be good to add this to the `.gitignore` so that it never gets accidentally committed.

MikeInnes:
:+1: thanks!

PR #29, opened 2017-10-03
minimal GPUArray integration

SimonDanisch:


SimonDanisch:
bump

SimonDanisch:
bump - any comments?

SimonDanisch:
bump

MikeInnes:
Pretty busy but will try to take a look this week. I need to go over all the mechanisms you're buying CuArrays into here, e.g. the functionality the abstract type gives me and the generic kernel mechanisms.

SimonDanisch:
bump! Would be nice to get this done sooner than later, since we already get into duplicated work for bug fixing - and I don't think I can fully announce GPUArrays without this.
It shouldn't take that much time, since 50% of there kernels are already from CuArrays and there are only like ~20. I'm willing to also adapt the broadcast a bit, since you seem to overload the Base interface more gracefully.

MikeInnes:
Sorry for the wait. The BLR gpu machine is down right now but I'll definitely pick out some time for this next week. I might still want to make the patch a bit smaller and take things piece by piece, but this PR will be a really useful starting point for that.

One thing I've been thinking about is whether it's actually right to have a `GPUArray` abstract type here, or if that's the wrong way to express that. The problem is that an abstract type can only express properties along one dimension, and Base has chosen to give properties of the data layout (e.g. `DenseArray`), so using it for GPU-ness prevents us from working with that. On top of which, there's no way to dispatch on `CPUArray` right now in a way that's consistent. An alternative might be to have something like a trait, which defaults to CPU but can also be GPU, distributed etc.

Just a thought, doesn't have to block this but seems worth discussing.

MikeInnes:
I've avoided implementing `pointer` so far because it seems easy to make mistakes with; people almost always use it to mean "cpu pointer" and it might be safer to have a separate `GPUArrays.pointer` function that's specific to GPU memory. Any thoughts?

src/blas/highlevel.jl
```diff
@@ -10,8 +10,17 @@ CublasArray{T<:CublasFloat} = CuArray{T}
 #
 ###########
 
-Base.scale!(x::CuArray{T}, k::Number) where T<:CublasFloat =
-  scal!(length(x), convert(eltype(x), k), x, 1)
+function Base.scale!(X::CuArray{T}, s::Real) where T <: Base.BLAS.BlasComplex
```
MikeInnes: Why is this needed? cublas already has methods for complex numbers

SimonDanisch: Ambiguities! But I will double check.

src/gpuarray_interface.jl
```diff
@@ -0,0 +1,110 @@
+import GPUArrays
+
+Base.similar(::Type{<: CuArray}, ::Type{T}, size::Base.Dims{N}) where {T, N} = CuArray{T, N}(size)
```
MikeInnes: This doesn't seem to have an analogue for other array types?

src/gpuarray_interface.jl
```diff
@@ -0,0 +1,110 @@
...
+
+#Abstract GPU interface
+immutable CUKernelState end
```
MikeInnes: should probably use `Cu` consistently

test/runtests.jl
```diff
@@ -21,7 +21,29 @@ function testf(f, xs...)
   @test collect(f(cu.(xs)...)) ≈ collect(f(xs...))
 end
 
+
+using Base.Test, GPUArrays.TestSuite
+using GPUArrays: JLArray
```
MikeInnes: What's this for?

MikeInnes:
RE types, given that GPUArray is a DenseArray, what's your plan for when we have support for strided and sparse arrays that aren't dense?

src/gpuarray_interface.jl
SimonDanisch: When you want to generically offer constructors like CuArray(rand(10)) you only have a type and no instance.
This solves that problem ;) If it was up to me, I'd love to have this in Base being the only method to overload!

SimonDanisch:
I haven't thought much about it, but I've always imagined them as a wrapper type, which doesn't inherit necessarily from GPUArrays:
```
 # AbstractSparseArray or whatever GPU equivalent we come up with in a sparse GPU Array package
struct SpareseArray <: AbstractSparseArray
values::GPUArray
indices::GPUArray
end
```

MikeInnes:
Doesn't that defeat the point of having a `GPUArray` type that you can dispatch on, write kernels for etc? You go back to needing a trait, but then you would have two inconsistent mechanisms for doing the same thing.

SimonDanisch:
True! Well, we can always add a node to the type hieararchy. I'd love to have a trait based approach, but currently we just don't have very nice tools for that.
Trait like approaches still take me back to c where you need to pass the length of a vector as a separate argument, making it impossible to treat an array as a single object ;)

src/blas/highlevel.jl
MikeInnes: Even with the ambiguities it makes sense to use CUBLAS's method for complex numbers rather than reimplementing it.

MikeInnes: The ambiguity here is caused by GPUArrays wrapping the method for Real rather than Number. So I have to explicitly override that in order to opt-out I guess, which is a bit nasty.

SimonDanisch: Okay lets fix that then ;)

MikeInnes:
I still have concerns about this. It's heavily specialised for one specific kind of array, and it's not clear to me that it's going to make it easier, rather than harder, to share code once we start having new or custom array types on the gpu. The one thing it does provide – the `GPUArray` interface – is not really reliable since it won't work with constructs like `RowVector`, and it seems pretty short-sighted to have no plan for sparse or strided arrays.

It also still bothers me that it's not clear what this buys CuArrays into; you say there's some 20 kernels it adds, but there's no good way to verify that other than weeding through 3,000 lines of GPUArrays.jl code. As we discussed, an opt-in approach like the test suite has would easily resolve this and other issues. As it is I hope you're going to be around to resolve any issues that do come up.

For now I guess we'll just have to try it and see what happens.

src/blas/highlevel.jl
SimonDanisch: https://github.com/JuliaGPU/GPUArrays.jl/pull/83

SimonDanisch:
Thanks mike! Well, it's the first design iteration, and I tried just to concentrate on sharing code between cuda and opencl for dense buffers. I hope to get you on board and in the future design things together! My main hope is, that we can now solve most problems inside GPUArrays so that improvements come free to all specialized array packages! We should get issues started for strided/sparse arrrays to figure out what would be needed for a better design! I think we also need a better story for e.g. adapt, views and device arrays!

SimonDanisch:
Would it help if I port the nicer broadcast from CuArrays to GPUArrays?
With promote_container_type & generated kernel, while keeping the faster gpu_sub2ind...
I could do this today, but then we should move the implementation out of CuArrays ;)

MikeInnes:
Sure, that seems like a good next step.

PR #36, opened 2017-11-01
Adding orgqr to return Q

jdnz:


src/solver/dense.jl
```diff
@@ -279,8 +279,9 @@ for (bname, fname, elty) in ((:cusolverDnSorgqr_bufferSize, :cusolverDnSorgqr, :
                                Cint, Ptr{$elty}, Ptr{$elty}, Cint, Ptr{Cint}),
                               cusolverDnhandle[1], m, n, k, A,
                               lda, tau, buffer, bufSize[], devinfo))
-            if devinfo[1] < 0
-                throw(ArgumentError("The $(devinfo[1])th parameter is wrong"))
+            info = collect(devinfo)
```
MikeInnes: how large is the `info` array? we could work around the `allowscalar` issue with an `unsafe_getindex` call or something, which might be better if info is large

jdnz: As I understand it is a single integer stored on the gpu that gives the success of the algorithm. Here we make a CuArray for that, but is there some way to just point to a single integer?

MikeInnes: Ah I see. Yeah that's fine then. We could theoretically have something like `CuRef` but I don't think it's worth it for this case.

MikeInnes:
Awesome, thanks a lot for the code cleanup as well. Is this ready to go?

jdnz:
It works at least! One question I have is that here, the cuSolver code allocates ```buffer  = CuArray(zeros($elty, bufSize[]))``` which is presumably non-optimal since it first creates a cpu zeros array. What is the best CuArray way to do this?

MikeInnes:
Yes, very much so. One option is to do `buffer = CuArray{Float32}(5).=0`.

Ideally we'd have a method like `zeros(CuArray{T}, dims...)`. Feel free to implement that if you want.

src/solver/dense.jl
```diff
@@ -256,6 +256,42 @@ for (bname, fname,elty) in ((:cusolverDnSgeqrf_bufferSize, :cusolverDnSormqr, :F
...
+            m = size(A , 1)
+            n = min(m, size(A, 2))
+            lda = max(1, stride(A, 2))
+            k = length(tau)
+            bufSize = Ref{Cint}(0)
+            statuscheck(ccall(($(string(bname)),libcusolver), cusolverStatus_t,
+                               (cusolverDnHandle_t, Cint, Cint, Cint, Ptr{$elty}, Cint,
+                                Ptr{$elty}, Ref{Cint}),
+                               cusolverDnhandle[1], m, n, k, A, lda, tau, bufSize))
+            buffer  = CuArray(zeros($elty, bufSize[]))
```
vchuravy: `CuArray{$elty}(bufSize[])` will create an uninitialised array on the GPU.

src/solver/dense.jl
```diff
@@ -256,6 +256,42 @@ for (bname, fname,elty) in ((:cusolverDnSgeqrf_bufferSize, :cusolverDnSormqr, :F
...
+            devinfo = CuArray(zeros(Cint, 1))
+            statuscheck(ccall(($(string(fname)), libcusolver), cusolverStatus_t,
+                              (cusolverDnHandle_t, Cint, Cint, Cint, Ptr{$elty},
+                               Cint, Ptr{$elty}, Ptr{$elty}, Cint, Ptr{Cint}),
+                              cusolverDnhandle[1], m, n, k, A,
+                              lda, tau, buffer, bufSize[], devinfo))
+            info = collect(devinfo)
```
vchuravy: `info = _getindex(devinfo, 1)`

jdnz:
I think I'm done working on this for the minute, unless you can see some improvements. As far as performance goes my initial tests seem to suggest that these routines are slower than the julia cpu ones by 1-2x. I don't know whether that is due to cusolver or the julia wrapping, or perhaps just because I'm only using a GeForce GT 750M.

vchuravy:
Can you post the script that you used for measuring and I can run it on
different hardware.

jdnz:
I wasn't doing anything sophisticated, just some simple tests with ```@time``` of the following functions.
```
import CuArrays.CUSOLVER.qonly!
function qonly!(b::Array)
    f = qrfact!(b)
    Array(f[:Q])
end
function qonly_swap!(b::CuArray)
    bcpu = collect(b)
    f = qrfact!(bcpu)
    copy!(b, Array(f[:Q]))
end

a = rand(Float32, 10000, 1000);
A = CuArray(a);
B = CuArray(a);
@time q = qonly!(A);
@time p = qonly!(a);
@time k = qonly_swap!(B);
```
After running ```@time``` a few times for that size of array my cpu took ~1s while the gpu ~2s.

MikeInnes:
Anyone want to review this before I merge? Perhaps @kshyatt?

test/solver.jl
```diff
@@ -142,8 +187,42 @@ k = 1
...
+        h_q            = collect(d_q)
+        q              = qrfact!(A)
+        @test h_q ≈ Array(q[:Q])
+        A              = rand(elty, n, m)
+        d_A            = CuArray(A)
+        d_q            = qrq!(d_A)
+        h_q            = collect(d_q)
+        q              = qrfact!(A)
+        @test h_q ≈ Array(q[:Q])
+        CuArrays.allowscalar(true)
```
MikeInnes: Why is this needed here? We shouldn't need this for any "supported" uses of CuArrays.

jdnz: The problem here is that the ```geqrf!``` function from cusolver returns a compact representation ```QR``` of the qr decomposition, with the ```R``` part being the upper triangular part of ```QR```. In making the high level implementation I just adjusted the qr implementation from julia base, where to return ```R``` they apply ```triu!``` which uses scalar indexing.

MikeInnes:
Ok, let's go ahead with this.

Thanks a lot @jdnz!

PR #40, opened 2017-12-11
Fixes and improvements

maleadt:
I started looking into replacing CUDAdrv.CuArray with CuArrays.jl; making it up to date with CUDAdrv 0.7 is the first step.

There seem to be some reliability issues though, broadcasting with `sin` doesn't seem to work (even before 8e555f77e187b678896364c95ead37dcc9f7d94c), and I am running into errors with both the CUBLAS and CUSOLVER integrations...

Relies on JuliaGPU/CUDAdrv.jl#71


maleadt:
So this is turning out quite a bit larger than anticipated, without well-defined scope. I should have broken this down into smaller, review-able pieces, but unless somebody really wants me to I'd prefer just to get everything working again and go from there. I promise to be a better CuArrays dev in 2018 😉

MikeInnes:
Seems totally fine to me, the commits are clear enough. At this point getting things in better shape to build on is the right priority.

src/array.jl
```diff
@@ -22,7 +22,7 @@ function unsafe_free!(xs::CuArray)
   return
 end
 
-Base.cconvert(::Type{Ptr{T}}, x::CuArray{T}) where T = x.buf
+Base.cconvert(::Type{<:Ptr}, x::CuArray) = x.buf
```
MikeInnes: I can see that we'd want to allow `convert(Ptr, xs)`, but would we want things like `convert(Ptr{String}, ::CuArray{Int})` to go through?

I guess as this is not that user-facing it may not matter.

maleadt: I didn't want to, but it currently the `cudnn` wrappers pass a CuArray to `ccall` expecting `Ptr{Void}`:
```julia
function cudnnSoftmaxForward(handle,algo,mode,alpha,xDesc,x,beta,yDesc,y)
    @check ccall((:cudnnSoftmaxForward,libcudnn),cudnnStatus_t,(cudnnHandle_t,cudnnSoftmaxAlgorithm_t,cudnnSoftmaxMode_t,Ptr{Void},cudnnTensorDescriptor_t,Ptr{Void},Ptr{Void},cudnnTensorDescriptor_t,Ptr{Void}),handle,algo,mode,alpha,xDesc,x,beta,yDesc,y)
end

function cudnnSoftmaxForward(src::CuArray, dest::CuArray=src;
                             handle=cudnnHandle,
                             algorithm=CUDNN_SOFTMAX_ACCURATE, # or CUDNN_SOFTMAX_FAST
                             mode=CUDNN_SOFTMAX_MODE_INSTANCE, # or CUDNN_SOFTMAX_MODE_CHANNEL
                             alpha=1.0, beta=0.0)
    cudnnSoftmaxForward(handle, algorithm, mode,
                        cptr(alpha, src), TD(src,4), src,
                        cptr(beta, dest), TD(dest,4), dest)
    return dest
end
```

Either we type the inner `cudnnSoftmaxForward`, or manually convert to a typed pointer in the outer one, neither seems nice to me. Other suggestions?

MikeInnes: Will plain `Ptr` work in the ccall? If not, perhaps we can just make `convert(Ptr{Void}, xs)` a special case that ignores eltype, which aligns well with the meaning in C anyway.

maleadt: Plain `Ptr` doesn't work with `unsafe_convert`, but the special case rule is a better solution indeed.

maleadt:
OK, tests consistently succeed here now. Let's wait until CI works again (needs a tag of CUDAnative, JuliaLang/METADATA.jl#12790) and merge afterwards. I'll have a look at 0.7 compatibility then.

PR #43, opened 2018-01-05
Fix build with CUDAapi 0.3.0

iblis17:
and bump CUDAapi requirement

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,6 +1,6 @@
 julia 0.6
 CUDAnative 0.5
 CUDAdrv 0.5
-CUDAapi
+CUDAapi 0.3.0
 NNlib
 GPUArrays
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -5,11 +5,11 @@ config = Dict()
 
 toolkit = CUDAapi.find_toolkit()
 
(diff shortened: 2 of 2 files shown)


iblis17:
well, I want set of version number of `CUDA*`/`Cu*` which works together.
(just doing some works with Flux, but frustrated with the compatibility problem

iblis17:
I'm on Julia `0.6.3-pre.1`, with CUDA 9.1.
I tried lots of combination, still cannot get `CuArray` work, I give up now.
Also notice a breaking change: from CUDAdrv 0.5.2 -> 0.5.3. It causes CUDAapi 0.2.1 build fail, due to the `debug` macro got deleted.

seems CUDA 9 + Julia 0.6 is infeasible.

MikeInnes:
Yeah, I think we need to wait for 0.7 to fully support all of the new stuff.

maleadt:
Sorry, I've been changing quite some stuff. Hadn't seen your PR, the fix is already included in #40.

maleadt:
> seems CUDA 9 + Julia 0.6 is infeasible.

Missed this bit. Yes, sadly this combination is impossible. CUDA 9 contains breaking changes we cannot generate code for with Julia 0.6; Julia 0.7 contains a patch making it possible again.

PR #50, opened 2018-01-27
integrate CUFFT

RalphAS:
This provides the basic functionality (so, e.g., Simon's PDE demos work with CuArrays). Some internals (names and error handling) differ from conventions used for other libraries; I can patch these in a few days if you care.


MikeInnes:
Looks great, thanks! I've tweaked the build step to be a bit more robust (in particular to work on my CUDA9 system), check that looks alright to you.

Also, would be nice if this caught a few more cases, like `fft(cu([1,2,3]))`. Probably best to have all numeric arrays go through CUFFT by default, with a conversion if need be.

RalphAS:
Your tweak looks fine.
I've tried to handle the conversion cases you mentioned. I don't know what to make of the git error in CI testing.

maleadt:
That sometimes happens when force pushing branches. I've restarted the build.

RalphAS:
I'm getting failures when trying to make a `CuArray{Float32}` from `CuArray{Int}`. This used to work (for `T=Int`):

```julia
realfloat(x::CuArray{T}) where {T<:Real} = copy1(typeof(cufftfloat(zero(T))), x)

function copy1(::Type{T}, x) where T
    y = CuArray{T}(map(length, indices(x)))
    copy!(y, x)
end
```
but now it calls an index-based version of `copy!` which is not compatible with `CuArrays`.
Is this a bug, or am I doing this the wrong way?

RalphAS:
@maleadt Could you restart the new CI build again? Thanks.

maleadt:
A close/reopen should also trigger a new build.

RalphAS:
CI seems satisfied with my part, but has trouble with NNLIB?

maleadt:
The nnlib failure seems awfully reproducible, and doesn't happen on master.
I'll have a look on the CI machine.

maleadt:
I really can't reproduce this. Went as far as running on the CI server, under the same user, using the same Julia and packages as checked out by buildbot...

MikeInnes:
I've tested this locally and it's all good. Great work!

PR #51, opened 2018-01-30
Enable array slices

MikeInnes:
@maleadt this could do with a quick look-over, is the approach reasonable?

### DIFF
--- a/src/array.jl
+++ b/src/array.jl
@@ -4,14 +4,17 @@ import CUDAnative: DevicePtr
 
 mutable struct CuArray{T,N} <: GPUArray{T,N}
   buf::Mem.Buffer
+  offset::Int
   dims::NTuple{N,Int}
-  function CuArray{T,N}(buf::Mem.Buffer, dims::NTuple{N,Integer}) where {T,N}
-    xs = new{T,N}(buf, dims)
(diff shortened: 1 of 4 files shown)


src/array.jl
```diff
@@ -22,8 +25,11 @@ function unsafe_free!(xs::CuArray)
   return
 end
 
-Base.cconvert(::Type{Ptr{T}}, x::CuArray{T}) where T = x.buf
-Base.cconvert(::Type{Ptr{Void}}, x::CuArray) = x.buf
+unsafe_buffer(xs::CuArray) =
+  Mem.Buffer(xs.buf.ptr+xs.offset, sizeof(xs), xs.buf.ctx)
```
maleadt: How do you make sure `offset+sizeof(xs)` never exceeds the original buffer, and goes out of bounds?
You could also use `Mem.view(::Buffer, offset)` for this, which does a bounds check on the new buffer.

MikeInnes: If that was the case then the original CuArray object would have been invalid to begin with. None of our constructors should allow this to happen.

MikeInnes: Although I think I still need to put in a bounds check in one place

PR #54, opened 2018-02-13
Support for Conv3D layer

tejank10:
Added CuArrays support for Conv3D


MikeInnes:
x-ref https://github.com/FluxML/NNlib.jl/pull/27

tejank10:
Is it failing because conv3d is not yet added in NNlib?

MikeInnes:
Yeah. You could `Pkg.checkout("NNlib")` in the build script in principle, but it's probably easier just to wait til that's merged.

PR #63, opened 2018-03-29
Move CURAND into CuArrays

### LINKED ISSUE #9: Native rand
Currently I have to do `cu(rand(...))`. It would be nice to have a native version.

andreasnoack:
Fixes #9. Tests pass but there might be some more cleanup to do.


src/rand/wrappers.jl
```diff
@@ -0,0 +1,212 @@
...
+           "CURAND_STATUS_INITIALIZATION_FAILED",
+         CURAND_STATUS_ARCH_MISMATCH => "CURAND_STATUS_ARCH_MISMATCH",
+         CURAND_STATUS_INTERNAL_ERROR => "CURAND_STATUS_INTERNAL_ERROR")
+
+function error_message(status::UInt32)
+    return "cuRAND operation failed with status: " * MESSAGE_BY_STATUS[status]
+end
+
+
+function statuscheck(status::UInt32)
```
maleadt: we've moved to a more sane macro+exception based error handling with the other wrappers, see eg. https://github.com/JuliaGPU/CuArrays.jl/blob/master/src/fft/error.jl

src/rand/core.jl
```diff
@@ -0,0 +1,9 @@
+
+using CUDAdrv
+ 
+include("init.jl")
+include("defs.jl")
+include("wrappers.jl")
+include("hl.jl")
+
+# println("At the end of module loading: $_rng")
```
maleadt: ?

src/rand/hl.jl
```diff
@@ -0,0 +1,49 @@
+# uniform
```
maleadt: maybe rename to `highlevel.jl` as with the BLAS wrapper

maleadt:
some minor comments. maybe it would make sense to have clang.jl-based wrappers for the api, such that at least all functions are wrapper (cfr. cuBLAS).

src/rand/init.jl
```diff
@@ -0,0 +1,19 @@
+
+@static if is_windows()
```
maleadt: all this shouldn't be necessary. we have functionality in CUDAapi for this.

andreasnoack:
Thanks for the comments. I think I've managed to address them except for the Clang part. I think it should be okay to do later. Could you restart the tests?

maleadt:
> I think it should be okay to do later.

yeah definitely

> Could you restart the tests?

seems like the open/close trick worked. you can also just log in on the CI page, everybody with JuliaGPU membership has access to that.

PR #64, opened 2018-03-30
Some improvements to test coverage

andreasnoack:
Enable blas, fft, and solver tests and fix a few minor issues with the solver tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,12 @@
 # CuArrays
 
 [![][buildbot-julia06-img]][buildbot-julia06-url]
+[![codecov][codecov-img]][codecov-url]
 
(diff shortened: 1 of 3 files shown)


andreasnoack:
@maleadt It looks like the cudnn tests trigger an assertion error in LLVM

maleadt:
Yes, this has been a recurring problem: https://github.com/JuliaGPU/CuArrays.jl/pull/50#issuecomment-363354166
I wouldn't have merged that PR, because it essentially breaks CI...

maleadt:
Alternatively, you can guard the failing test behind a [`ENV["CONTINUOUS_INTEGRATION"]`](https://github.com/maleadt/juliagpu-buildbot/blob/397ad8b3bcbbd6e2dade1115201af753392f8e78/master.cfg#L321-L323) check

andreasnoack:
I think I got it now. The CUDNN tests just have to be run before any of blas, solver, and fft.

PR #65, opened 2018-03-31
Make cholfact(CuArray) work

andreasnoack:
Make potrf and potrs methods of the Base versions

Use some convenience functions from Base


andreasnoack:
The is a "scalar getindex is disabled" reported in the log. Is that a special setting for the build bot? I didn't see it on Cyclops.

vchuravy:
The tests set `CuArrays.allowscalar(false)` which prevents slow fallback routines. But you should see that locally as well if you use `test/runtests.jl`

andreasnoack:
Ah. I just pasted the tests I'd modified so `CuArrays.allowscalar(false)` hadn't been set. I guess I need to write some GPU code. We'd need `issymmetric` and is `ishermitian` on the GPU.

vchuravy:
bump! Would be good to have this.

PR #74, opened 2018-04-20
curand for Float32

vchuravy:


vchuravy:
Alan, also immediately tried `curand(Float32, 1024, 1024) and was sad that it didn't work ;)

andreasnoack:
I think we should clean up the API here such that it is as close to Base possible.

PR #83, opened 2018-05-29
Managed pool allocator

### LINKED ISSUE #79: GPU memory hog
The current pool allocator never frees memory except when allocations fail. This quickly leads to CuArrays/Flux using all of the available memory, as frequently observed on `cyclops`. We should probably have an idle thread scan for unused pool entries and free those up.

maleadt:
Improve the pooled memory allocator:
- keep track of both used and available buffers
- keep track of current pool usage
- add a background task to keep track of recent pool usage
- add a background task to reclaim memory from underused pools

Very ad-hoc, couldn't find much existing work on allocators that sit wrap another dynamic allocator (and still free memory), but seems to work pretty nicely.

Interestingly, this also improves performance, by not trashing the Julia GC, and avoiding release of other memory buffers during aggressive memory reclamation. For example, with MNIST:
- identical execution time
- gc time 4,7% -> 1,8%
- 861 allocs -> 576 allocs
- memory allocated: 5,7 GiB -> 3,9 GiB

Of course, the main use case is eg. MNIST + `sleep(10)` (or some idle REPL time), at which point 539 frees kick in and memory usage is reduced to about 600MiB.

Still contains some debugging print statements. Passes CuArrays test suite, but need to test and measure performance on e.g. the Flux test suite.

Fix #79 (cc @jekbradbury)

### DIFF
--- a/src/CuArrays.jl
+++ b/src/CuArrays.jl
@@ -5,6 +5,8 @@ module CuArrays
 using CUDAdrv, CUDAnative
 import CUDAnative: cudaconvert
 
+using MicroLogging
+
 export CuArray, CuVector, CuMatrix, CuVecOrMat, cu, cuzeros, cuones
 
 const ext = joinpath(dirname(@__DIR__), "deps", "ext.jl")
--- a/src/CuArrays.jl
+++ b/src/CuArrays.jl
@@ -42,6 +42,8 @@ function __init__()
         warn("Please run Pkg.build(\"CuArrays\") and restart Julia.")
         return
     end
+
+    __init_memory__()
 end
 
 end # module
--- a/src/CuArrays.jl
+++ b/src/CuArrays.jl
@@ -5,8 +5,6 @@ module CuArrays
 using CUDAdrv, CUDAnative
 import CUDAnative: cudaconvert
 
-using MicroLogging
-
 export CuArray, CuVector, CuMatrix, CuVecOrMat, cu, cuzeros, cuones
 
 const ext = joinpath(dirname(@__DIR__), "deps", "ext.jl")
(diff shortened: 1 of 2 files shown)


maleadt:
At this point, the allocator is tuned to release underused fractions of each pool once a part is unused for 5 ticks, where a tick is at least 1 second apart (when pools are changing), and at most 5 seconds (when we're in a stable regime). This works fine with the MNIST GPU examples, improving performance with the convolutional network, while significantly increasing allocations with the MLP demo but not hurting performance.

For some more data points, please run and code of yours with the `CUARRAYS_MANAGED_POOL` environment variable set to either `true` or `false`. This will disable the background pool management, and print pool statistics at Julia exit. Ideally, also wrap the time-sensitive part of your code with `@time` so that we get an idea about the GC time:

```
$ CUARRAYS_MANAGED_POOL=false jj mnist/conv.jl
loss(X, Y) = 2.3221996f0 (tracked)
loss(X, Y) = 0.32633972f0 (tracked)
 23.653550 seconds (6.35 M allocations: 334.923 MiB, 4.23% gc time)
req_alloc: 6053
req_free: 5686
actual_alloc: 854
actual_free: 0
amount_alloc: 6187093313
amount_free: 0
alloc types: 5192 854 7 0

$ CUARRAYS_MANAGED_POOL=true jj mnist/conv.jl
Managing pool
loss(X, Y) = 2.3221996f0 (tracked)
loss(X, Y) = 0.32144815f0 (tracked)
 23.483109 seconds (6.38 M allocations: 336.515 MiB, 1.22% gc time)
req_alloc: 6053
req_free: 5677
actual_alloc: 532
actual_free: 0
amount_alloc: 3962633089
amount_free: 0
alloc types: 5521 532 0 0
```

maleadt:
OK, let's then merge this to give it some exposure.

PR #84, opened 2018-06-06
Add GitLab CI.

maleadt:


maleadt:
Right, source build.

maleadt:
@MikeInnes I don't have sufficient permissions to do the same for Flux.jl, so I'll let you configure that.

PR #86, opened 2018-06-09
Dilation support for Conv

tejank10:
Dilation has been added in the latest release of NNlib. Because of that `NNlib.conv!` doesn't run on GPU since those changes have to be reflected here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 [![][codecov-img]][codecov-url]
 
 [gitlab-img]: https://gitlab.com/JuliaGPU/CuArrays.jl/badges/master/pipeline.svg
-[gitlab-url]: https://gitlab.com/JuliaGPU/CuArrays.jl/commits/master
+[gitlab-url]: https://gitlab.com/JuliaGPU/CuArrays.jl/pipelines
 
 [codecov-img]: https://codecov.io/gh/JuliaGPU/CuArrays.jl/branch/master/graph/badge.svg
 [codecov-url]: https://codecov.io/gh/JuliaGPU/CuArrays.jl
--- a/src/dnn/helpers.jl
+++ b/src/dnn/helpers.jl
@@ -92,12 +92,12 @@ end
 pdsize(w, nd)=Cint[reverse(psize(w,nd))...]
 psize(w, nd)=(isa(w,Integer)  ? fill(w,nd) : length(w) != nd ? error("Dimension mismatch") : w)
 
-function ConvDesc(T, N, padding, stride, upscale, mode)
+function ConvDesc(T, N, padding, stride, dilation, upscale, mode)
     cd = cudnnConvolutionDescriptor_t[0]
     cudnnCreateConvolutionDescriptor(cd)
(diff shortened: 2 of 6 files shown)


MikeInnes:
Great, can you add some simple tests?

tejank10:
I was testing it with current tests, which it failed. All six tests fail at [this](https://github.com/JuliaGPU/CuArrays.jl/blob/ad449f44bb740a3b3d8fde84046f3c26bf33996d/src/dnn/libcudnn.jl#L66) line in `libcudnn.jl` which is a `ccall`:
```
@check ccall((:cudnnGetConvolutionNdDescriptor,libcudnn),cudnnStatus_t,(cudnnConvolutionDescriptor_t,Cint,Ptr{Cint},Ptr{Cint},Ptr{Cint},Ptr{Cint},Ptr{cudnnConvolutionMode_t},Ptr{cudnnDataType_t}),convDesc,arrayLengthRequested,arrayLength,padA,strideA,upscaleA,mode,dataType)
```

MikeInnes:
Any idea when that changed? It'd be good to fix this but I'd rather not break older CUDNN versions.

As we're fixing this, maybe it's a good idea to have a quick patch that accepts the kwarg and throws an error if it's not the default value. Then basic convolutions will work at least.

MikeInnes:
Any update on this?

MikeInnes:
Bump -- would be great to get this in. What's the status at the moment?

tejank10:
First of all, apologies for a late reply.
From the current code, I presume we are using cuDNN v4. I just checked the [v6 release notes](https://developer.download.nvidia.com/compute/machine-learning/cudnn/secure/v6/prod/Doc/cuDNN_v6.0.21_ReleaseNotes.pdf?gF2mBftHA1zXEmC60RyEBwShk9kmfDK1hg_2jYwU7y0qcNMy8sTtmEwQ957zyVrDnY2MA2TT1KQC-YGo9TRAGKDXCaVVLMiyG7V_5CEz9VkxxRr7pLid6pYi4JeRwbD0t-Q-MZ6l-3VQLVkY_fvKlxMzxydCmeNJ2v-zuPS4PWq30rd_HZfYNMrlZXFXerE) which is version where dilatied convolution was introduced. It mentions replacing `upscale` with `dilation`.

> Dilated Convolutions: Dilated Convolutions are now supported in cuDNN without a 
change in API. Previously unused “upscale” fields in the Convolution Descriptor 
have been repurposed to allow user specification of dilation factors along each 
dimension. Support for dilation is present in the following code paths: 
Forward : CUDNN_CONVOLUTION_FWD_ALGO_IMPLICIT_GEMM, 
Backward Data: CUDNN_CONVOLUTION_BWD_DATA_ALGO_0 and 
Backward Filter: CUDNN_CONVOLUTION_BWD_FILTER_ALGO_0

Currently, cuDNN v7 is the latest one. Are we having any blocks with regard to moving to v7?

MikeInnes:
No problem at all, I'm just looking forward to having this in :)

This code is inherited from CUDNN 4 code, yeah. AFAIK the convs that we have are working on v7 as well though. What might work well is to define different wrappers depending on the CUDNN version installed (and you can throw an error if `dilation` is non-default, where CUDNN doesn't support it).

PR #100, opened 2018-07-21
Add support for Convolution Forward Bias and Other Activation Functions

avik-pal:


avik-pal:
All the new wrappers are functional. Only `identity` is not working for `cudnnConvolutionBiasActivationForward`. The other thing that remains is adding a few tests.

src/dnn/helpers.jl
```diff
@@ -115,3 +115,16 @@ function PoolDesc(nd, window, padding, stride, mode, maxpoolingNanOpt=CUDNN_NOT_
     finalizer(free, this)
     return this
 end
+
+mutable struct ActivationDesc; ptr; end
```
maleadt: Why is this mutable & untyped? It doesn't look like `ptr` is ever assigned to, let alone changing types.

src/dnn/helpers.jl
```diff
@@ -115,3 +115,16 @@ function PoolDesc(nd, window, padding, stride, mode, maxpoolingNanOpt=CUDNN_NOT_
...
+free(ad::ActivationDesc)=cudnnDestroyActivationDescriptor(ad.ptr)
+Base.unsafe_convert(::Type{cudnnActivationDescriptor_t}, ad::ActivationDesc)=ad.ptr
+
+function ActivationDesc(mode, coeff, reluNanOpt=CUDNN_NOT_PROPAGATE_NAN)
+    ad = cudnnActivationDescriptor_t[0]
```
maleadt: should be `Ref{cudnnActivationDescriptor_t}()`

src/dnn/helpers.jl
avik-pal: I was simply following the convention as in `ConvDesc` and `PoolDesc`.

maleadt: Yeah, I noticed. Should be improved, but can happen in a different PR. Do fix the `Ref` stuff though, because I recently went through and fixed existing code.

avik-pal:
```julia
julia>?CuArrays.CUDNN.cudnnAddTensor #Shows 2 functions
julia>CuArrays.CUDNN.cudnnAddTensor(rand(1,1,10,1), rand(10,10,10,1)) #Gives no method matching exception
```
Any reason why this is happening. Same for the other functions I added.

vchuravy:
`typeof(rand(1,1,10,1)) == Array{Float32, 4}` and `cudnnAddTensor(A::CuArray{T,N}, C::CuArray{T,N}`? You have to convert the `Array` to a `CuArray` first

PR #104, opened 2018-08-08
fix more deprecations + preprocess

SimonDanisch:


maleadt:
Contained in https://github.com/JuliaGPU/CuArrays.jl/pull/105

PR #105, opened 2018-08-08
Fixes for 0.7 and 1.0

maleadt:


maleadt:
Let's just merge this, even though a WIP it's better not to have these kind of huge PRs.

PR #106, opened 2018-08-11
Further improvements for 1.0

andreasnoack:
I've started looking into the linear algebra fixes.


andreasnoack:
There, the BLAS tests now pass. It requires a small fix to Adapt which I'll submit momentarily

maleadt:
https://github.com/JuliaGPU/GPUArrays.jl/pull/135

PR #108, opened 2018-08-15
Cleanup broadcast implementation and printing

vchuravy:
Fixes https://github.com/JuliaGPU/CUDAnative.jl/issues/223

- By using `print_array` instead of `show` we get the smarts of the base implementation.
- Fixes the `RefValue` handling by introducing an immutable wrapper
- fixes the `preprocess` hack by using style `Nothing`

### DIFF
--- a/src/array.jl
+++ b/src/array.jl
@@ -143,13 +143,9 @@ cuones(T::Type, dims...) = fill!(CuArray{T}(dims...), 1)
 cuzeros(dims...) = cuzeros(Float32, dims...)
 cuones(dims...) = cuones(Float32, dims...)
 
-Base.show(io::IO, x::CuArray) = show(io, collect(x))
-Base.show(io::IO, x::LinearAlgebra.Adjoint{<:Any,<:CuArray}) = show(io, LinearAlgebra.adjoint(collect(x.parent)))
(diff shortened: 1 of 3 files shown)


MikeInnes:
It looks like `preprocess` is doing what I imagined we'd do by overloading `Broadcast.broadcasted(::ArrayStyle{<:CuArray}, ...`. Is preprocess a Base thing, or how is this broadly working? If I understand this right and the intent is to fix broadcasting automatically, we can remove a lot of NNlib's cruft around it.

maleadt:
If this preprocessing works, we should revert https://github.com/JuliaGPU/CuArrays.jl/commit/d96f4e958e4e6a6dd305b516ac6e23f0eae84f38

vchuravy:
I don't quite remember why i went down the `Broadcast.preprocess` route over `Broadcast.broadcasted` and Slack has eaten my discussion with @mbauman about this. IIRC the main difference is that `preprocess` allows you to specialise on the output container whereas `broadcasted` uses the ArrayStyle.

I will try reverting d96f4e9 and see what I get.

MikeInnes:
One issue with `preprocess` is that it's not compatible with `Broadcast.flatten`, so you have to explicitly hook the preprocessing back in if you're using that.

PR #111, opened 2018-08-16
faster mapreducedim kernel

jekbradbury:
The algorithm here is to distribute 1024 threads/block among the inner
(reduction) and outer (map) dimensions; the remaining outer dimensions
get scheduled as the block grid. If fewer than 8 threads would be
dedicated to the inner dimension (because the outer dimension is much
larger) then we revert to the original kernel (which has the advantage
that it doesn't need to copy data to and from shared memory).

I'm now using static shared memory, so this is potentially portable to
GPUArrays. I don't think I'll get a chance to do that in the next couple
days, so I'd recommend merging it here first.

Benchmarks below (`serial` is the old kernel; `parallel` is the new kernel
with block-parallel reduction; `polyalgo` is the final version that chooses
between them):
```
from (2, 2) to (2, 1):
         CPU:   30.045 ns (0 allocations: 0 bytes)
  serial GPU:   13.876 μs (28 allocations: 864 bytes)
parallel GPU:   15.359 μs (31 allocations: 1.16 KiB)
polyalgo GPU:   14.004 μs (28 allocations: 864 bytes)
from (1024, 1024) to (1024, 1):
         CPU:  268.191 μs (0 allocations: 0 bytes)
  serial GPU:  216.220 μs (28 allocations: 864 bytes)
parallel GPU:   42.415 μs (39 allocations: 1.28 KiB)
polyalgo GPU:   42.238 μs (39 allocations: 1.28 KiB)
from (4096, 4096) to (4096, 1):
         CPU:   12.739 ms (0 allocations: 0 bytes)
  serial GPU:    1.497 ms (28 allocations: 864 bytes)
parallel GPU:  234.669 μs (39 allocations: 1.28 KiB)
polyalgo GPU:  234.497 μs (39 allocations: 1.28 KiB)
from (32768, 32768) to (32768, 1):
         CPU:    1.199  s (0 allocations: 0 bytes)
  serial GPU:   13.838 ms (28 allocations: 864 bytes)
parallel GPU:   13.669 ms (39 allocations: 1.28 KiB)
polyalgo GPU:   13.669 ms (39 allocations: 1.28 KiB)
from (1048576, 1024) to (1048576, 1):
         CPU:    1.200  s (0 allocations: 0 bytes)
  serial GPU:    5.452 ms (29 allocations: 880 bytes)
parallel GPU:   10.320 ms (40 allocations: 1.30 KiB)
polyalgo GPU:    5.451 ms (29 allocations: 880 bytes)
from (1024, 1048576) to (1024, 1):
         CPU:  692.394 ms (0 allocations: 0 bytes)
  serial GPU:  378.194 ms (28 allocations: 864 bytes)
parallel GPU:   12.487 ms (39 allocations: 1.28 KiB)
polyalgo GPU:   12.535 ms (39 allocations: 1.28 KiB)
from (32, 33554432) to (32, 1):
         CPU:  699.446 ms (0 allocations: 0 bytes)
  serial GPU:   11.362  s (28 allocations: 864 bytes)
parallel GPU:   25.778 ms (35 allocations: 1.22 KiB)
polyalgo GPU:   25.776 ms (35 allocations: 1.22 KiB)
from (33554432, 32) to (33554432, 1):
         CPU:    1.825  s (0 allocations: 0 bytes)
  serial GPU:    6.051 ms (29 allocations: 880 bytes)
parallel GPU:   11.234 ms (36 allocations: 1.23 KiB)
polyalgo GPU:    6.052 ms (29 allocations: 880 bytes)
```

### DIFF
--- a/src/mapreduce.jl
+++ b/src/mapreduce.jl
@@ -30,7 +30,7 @@ function mapreducedim_kernel_parallel(f, op, R::CuDeviceArray{T}, A::CuDeviceArr
         Ri = Ri_base + (blockIdx().x - 1) * blockDim().y + threadIdx().y
         Ri > Rlength && return
         RI = Tuple(CartesianIndices(R)[Ri])
-        S = @cuStaticSharedMem(T, 1024)
+        S = @cuStaticSharedMem(T, 512)
         Si_folded_base = (threadIdx().y - 1) * blockDim().x
         Si_folded = Si_folded_base + threadIdx().x
         # serial reduction of A into S by Slength ÷ xthreads
@@ -46,7 +46,7 @@ function mapreducedim_kernel_parallel(f, op, R::CuDeviceArray{T}, A::CuDeviceArr
             end
         end
         # block-parallel reduction of S to S[1] by xthreads
-        reduce_block(view(S, (Si_folded_base + 1):1024), op)
+        reduce_block(view(S, (Si_folded_base + 1):512), op)
         # reduce S[1] into R
         threadIdx().x == 1 && (R[Ri] = op(R[Ri], S[Si_folded]))
     end
@@ -56,8 +56,8 @@ function Base._mapreducedim!(f, op, R::CuArray{T}, A::CuArray{T}) where {T}
     Rlength = length(R)
     Ssize = ifelse.(size(R) .== 1, size(A), 1)
     Slength = prod(Ssize)
-    outer_thr = min(nextpow(2, Rlength ÷ 512 + 1), 1024)
-    inner_thr = min(1024 ÷ outer_thr, Slength)
+    outer_thr = min(nextpow(2, Rlength ÷ 512 + 1), 512)
+    inner_thr = min(512 ÷ outer_thr, Slength)
     if inner_thr < 8 # we can saturate the GPU with serial reduction
         range = ifelse.(length.(axes(R)) .== 1, axes(A), nothing)
         blk, thr = cudims(R)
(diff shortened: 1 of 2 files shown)


SimonDanisch:
If we can turn CuDynamicSharedMem into CuStaticMem, this could easily live in GPUArrays ;)

src/mapreduce.jl
```diff
@@ -0,0 +1,43 @@
...
+            arr[threadIdx().x] = op(arr[threadIdx().x],
+                                    arr[threadIdx().x + skip])
+        end
+        len = (len + 1) >> 1
+    end
+    sync_threads()
+end
+
+function mapreducedim_kernel(f::F, op::OP, R::CuDeviceArray{T},
+                             A::CuDeviceArray{T}, window) where {T, F, OP}
```
vchuravy: If you say:

```
::Val{window}), where {window, ...}
```

You can specialise the kernel and use static shared memory

jekbradbury:
This should be ready. It now uses static shared memory, so it could potentially be moved to GPUArrays, but would probably need to be reworked first (like the original mapreduce kernel) to be compatible with Transpiler.

MikeInnes:
Awesome work. Does one of those old version benchmarks really say 11 seconds? Either way, this is clearly a great improvement.

I imagine we are testing reduction already, but it'd be useful to test this kernel is getting called rather than the generic one. I'm not immediately sure how best to do that though.

jekbradbury:
Yep, 11 seconds comes from a linear, serial scan over 33 million elements. I’m also unsure how to test that a particular kernel is being called—would be straightforward with Cassette but that’s probably not an acceptable test dependency.

maleadt:
Issue for the test functionality: https://github.com/JuliaGPU/GPUArrays.jl/issues/152

skariel:
I'm having the "too many resources" problem, just submitted a new issue here: #131

PR #117, opened 2018-08-17
Fix bug in mul!

andreasnoack:
~~Increase mul! coverage in tests~~

Restructure BLAS tests a bit to make it easier to locate test failures

### DIFF
--- a/src/blas/highlevel.jl
+++ b/src/blas/highlevel.jl
@@ -115,8 +115,7 @@ function gemm_wrapper!(C::CuVecOrMat{T}, tA::Char, tB::Char,
 end
 
 # Mutating
-LinearAlgebra.lmul!(C::CuMatrix{T}, A::CuMatrix{T}, B::CuMatrix{T}) where T<:CublasFloat =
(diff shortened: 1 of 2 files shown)


maleadt:
Fold into https://github.com/JuliaGPU/GPUArrays.jl/blob/master/src/testsuite/blas.jl? Not sure which part of the API is unique to CuArrays.BLAS though.

andreasnoack:
It looks like a natural home for these tests. I'll move that part over but leave the bugfix and reformatting in this PR.

andreasnoack:
I've opened https://github.com/JuliaGPU/GPUArrays.jl/pull/143 which should cover the same as here. I'll update this PR once we have a new release of GPUArrays.

andreasnoack:
The tests revealed that a few methods were missing. This still depends on https://github.com/JuliaGPU/GPUArrays.jl/pull/144 and https://github.com/JuliaGPU/CuArrays.jl/pull/118 accurate so we should restart the tests once they are merged and we have a new release of GPUArrays.

PR #118, opened 2018-08-18
Remove redundant getindex and setindex! methods covered by GPUArrays methods

andreasnoack:
Fixes an issue with `allowscalar` not having an effect for `CuArrays`.


andreasnoack:
Okay. The test failure here is expected. It's the combination of the improved coverage of the tests in GPUArrays and that `allowscalar` working again. I've verified that the tests pass when combining this with #117 so I'll merge here, restart the tests in #117 and merge once they pass.

PR #119, opened 2018-08-20
Use `broadcasted` instead of `preprocess`

MikeInnes:
I'm not sure what the motivation for using `preprocess` was, but the current implementation does not actually work (e.g. `log.(xs) .+ 1`). Aside from that it means that the fix is not compatible with `Broadcast.flatten`, which is useful for many things.


MikeInnes:
Merging so I can finally get a working Flux release out. We can discuss changing the implementation later if there are strong reasons to do so.

PR #121, opened 2018-08-22
Fallback generic matmul

maleadt:


maleadt:
@andreasnoack Base uses different loops depending on the layout of the matrix: https://github.com/JuliaLang/julia/blob/d55b04430c9eebc229b6dafd8725697d86acf1b3/stdlib/LinearAlgebra/src/matmul.jl#L658-L745

Aren't the lazy wrappers made for that, as I do in this PR? That is, not passing `.parent` and switching the matmul computation, but just pass the wrapped array and do the computation on that? I mean, I can imagine optimized implementations for each type of layout, but that doesn't seem to be the case in Base.

andreasnoack:
I'll take a closer look at this later.

maleadt:
I'm going to merge this already because it blocks some other work, we can always improve it later.

PR #135, opened 2018-09-18
Don't return from kernels.

maleadt:


maleadt:
Interestingly, this shouldn't have caused test failures since the generic matmul isn't tested by the GPUARrays test suite. This means we're, wrongfully, dispatching to these implementations instead of using CUBLAS. Ref https://github.com/JuliaGPU/GPUArrays.jl/issues/152

maleadt:
Ref https://github.com/JuliaGPU/CUDAnative.jl/pull/254

PR #144, opened 2018-09-23
Rebase `Make cholfact(CuArray) work`

vchuravy:


vchuravy:
Ok this should work now in conjuncture with https://github.com/JuliaGPU/GPUArrays.jl/pull/162 (which tested this branch as well)

maleadt:
Restarted CI now that GPUArrays with ishermitian is tagged, but still triggers `scalar getindex is disallowed`.

vchuravy:
It was still testing against GPUArrays `0.4.0`

PR #150, opened 2018-09-25
Clean-ups (NFC).

maleadt:
Also serves to CI-test Julia master support on new CUDAnative tag.


maleadt:
Yay, works on `master` again.

PR #164, opened 2018-10-23
Adapt to changes in Adapt.jl

maleadt:
... and related CUDAnative changes


maleadt:
This also implies dropping support for Julia 0.7

PR #168, opened 2018-10-25
Accurately keep track of library handles

maleadt:
Two goals: 1) multi-device means multiple handles, 2) eagerly instantiating handles consumes a lot of memory (I've seen 200MB per CUDNN handle).

Didn't rework CUFFT because the code is a bit of a mess and it looks like the plans are not instantiated eagerly/globally?

Finally, destroying the CURAND handle crashes, so I've disabled it. Maybe it's not worth to ever call these methods since we're exiting the process anyway...

```
signal (11): Segmentation fault
in expression starting at no file:0
curandDestroyGenerator at /home/tbesard/CUDA/toolkit/current/lib64/libcurand.so (unknown line)
macro expansion at /home/tbesard/Julia/CuArrays/src/rand/error.jl:49 [inlined]
destroy_generator at /home/tbesard/Julia/CuArrays/src/rand/libcurand.jl:11
unknown function (ip: 0x7f746cee6040)
jl_fptr_trampoline at /buildworker/worker/package_linux64/build/src/gf.c:1831
jl_apply_generic at /buildworker/worker/package_linux64/build/src/gf.c:2184
jl_apply at /buildworker/worker/package_linux64/build/src/julia.h:1537 [inlined]
run_finalizer at /buildworker/worker/package_linux64/build/src/gc.c:115
jl_gc_run_finalizers_in_list at /buildworker/worker/package_linux64/build/src/gc.c:210
jl_gc_run_all_finalizers at /buildworker/worker/package_linux64/build/src/gc.c:245
jl_atexit_hook at /buildworker/worker/package_linux64/build/src/init.c:260
main at /buildworker/worker/package_linux64/build/ui/repl.c:234
__libc_start_main at /usr/lib/libc.so.6 (unknown line)
_start at /home/tbesard/Julia/julia-1.0/build/binary/bin/julia (unknown line)
Allocations: 547321220 (Pool: 547216498; Big: 104722); GC: 1286
```

### DIFF
--- a/src/CuArrays.jl
+++ b/src/CuArrays.jl
@@ -59,6 +59,7 @@ function __init__()
 
         # wipe the active handles
         isdefined(CuArrays, :CUBLAS) && (CUBLAS._handle[] = C_NULL)
+        isdefined(CuArrays, :CURAND) && (CURAND._generator[] = nothing)
         isdefined(CuArrays, :CUDNN)  && (CUDNN._handle[] = C_NULL)
     end
     push!(CUDAnative.device!_listeners, callback)
--- a/src/CuArrays.jl
+++ b/src/CuArrays.jl
@@ -34,6 +34,10 @@ include("mapreduce.jl")
 
 include("gpuarray_interface.jl")
 
+# many libraries need to be initialized per-device (per-context, really, but we assume users
+# of CuArrays and/or CUDAnative only use a single context), so keep track of the active one.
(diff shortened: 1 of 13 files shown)


maleadt:
Will need to adapt Flux.jl since the package uses CUDNN handles explicitly. Another reason to have that code here (ref https://github.com/FluxML/Flux.jl/pull/445#issuecomment-432605547).

PR #170, opened 2018-10-26
Sparse matrix support

kshyatt:
* Pull in CUSPARSE.jl, mostly modernized for Julia 1.0
* Pull in CUSOLVER.jl Sparse solvers, also modernized
* Tests pass
* Some improvements can be made, of course
  - some tests fail if uncommented
  - the style ain't great, friends (type horrorshow)
  - Doesn't incorporate new `lmul`/`rmul` LinAlg stuff ( will make a commit for this )


kshyatt:
Updated the PR to resolve the merge conflict.

kshyatt:
What's up with all those GPUArrays errors?

maleadt:
Looks like the scalar indexing assertion triggers when showing the arrays that failed a `@test`? We've only implemented functions of eg. `show_delim_array` that flip the `allowscalar` switch [for some types](https://github.com/JuliaGPU/GPUArrays.jl/blob/71c4621d1151a816f6f27e4b0259ba5c8e092773/src/abstractarray.jl#L53-L74), not all wrappers like `::LowerTriangular{Float32,CuArray{Float32,2}}`.

maleadt:
Added some unwrappers in https://github.com/JuliaGPU/GPUArrays.jl/pull/170. Branch has the same name over there, so CuArrays here should use it and hopefully show better errors now.

maleadt:
OK, that looks much better: the show errors are gone, just some ambiguities with CUBLAS now.

src/sparse/highlevel.jl
```diff
@@ -0,0 +1,30 @@
+(\)(A::AbstractTriangular{<:CuSparseMatrix},B::CuMatrix)       = sm('N',A,B,'O')
+(\)(transA::Transpose{<:Any, AbstractTriangular{<:CuSparseMatrix}}, B::CuMatrix) = sm('T',transA.parent,B,'O')
+(\)(adjA::Adjoint{<:Any, AbstractTriangular{<:CuSparseMatrix}},B::CuMatrix) = sm('C',adjA.parent,B,'O')
+
+(*)(A::CuSparseMatrix,B::CuVector)       = mv('N',A,B,'O')
```
maleadt: What's the reason you chose to implement the non-mutating versions, instead of directly using `mv!` and `mm2!`?

maleadt: I'm not entirely sure either, but I figured implementing the in-place ops is more versatile since there's asbtract functionality in Base for allocating containers with non-mutating ops, eg. https://github.com/JuliaLang/julia/blob/eabc5de03131e780a8be3b58dd576f1007b9ce99/stdlib/LinearAlgebra/src/matmul.jl#L140-L143

maleadt:
Let's merge this, we can switch to in-place ops later without breaking code.

PR #172, opened 2018-10-29
Simplify array constructors.

maleadt:
CuArrays/GPUArrays currently define lots of constructors (and methods of `similar` and `convert`) that have diverged from Base and really should not exist. This PR tries to get rid of them, to improve maintainability and code portability.

As part of that work, I propose to move essential functionality such as constructors to the concrete packages (ie. CuArrays). Defining these methods in a generic fashion is cumbersome, introduces loads of ambiguities, and results in applicable but nonsensical method invocations. For example:

```julia
julia> convert(GPUArray, cu([1]))
ERROR: MethodError: no method matching GPUArray{Float32,N} where N(::UndefInitializer, ::Tuple{Int64})
Stacktrace:
 [1] similar(::Type{GPUArray{Float32,N} where N}, ::Tuple{Int64}) at ./abstractarray.jl:618
 [2] convert(::Type{GPUArray}, ::CuArray{Float32,1}) at /home/tbesard/Julia/GPUArrays/src/construction.jl:97
 [3] top-level scope at none:0
```

This because of a generic `Base.convert(::Type{<: GPUArray{T1}}, ...)` that is only defined to be invoked on concrete subtypes. Base doesn't attempt to share such functionality between Array and AbstractArray.

Obviously massively breaking, but seems to pass tests except for some CUSOLVER failure.

### DIFF
--- a/src/array.jl
+++ b/src/array.jl
@@ -56,41 +56,26 @@ Base.sizeof(x::CuArray) = Base.elsize(x) * length(x)
 
 ## interop with other arrays
 
-# type and dimensionality specified
-CuArray{T,N}(x::AbstractArray{S,N}) where {T,N,S} = convert(CuArray{T,N}, x)
-
-# type but not dimensionality specified
-CuArray{T}(A::AbstractArray{S,N}) where {T,N,S} = CuArray{T,N}(A)
-
-# dimensionality but not type specified
-(::Type{CuArray{T,N} where T})(x::AbstractArray{S,N}) where {S,N} = CuArray{S,N}(x)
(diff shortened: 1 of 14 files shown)


andreasnoack:
Hm. Master is currently failing a ton of tests for me and I suspect that this PR is involved but it looks like CI went through just fine.

maleadt:
GPUArrays from master? The current `runtests.jl` matches branches on CI.

andreasnoack:
No that was the problem. I'd only checked out `CuArrays` master.

PR #175, opened 2018-10-30
Fixed SVD test

mbeltagy:
The U Matrix was not being tested properly. CPU matrix was not called in one instance and not parametrized in another.

### DIFF
--- a/test/solver.jl
+++ b/test/solver.jl
@@ -179,7 +179,7 @@ k = 1
         h_U            = collect(d_U)
         h_Vt           = collect(d_Vt)
         F              = svd(A, full=true)
-        @test abs.(h_U'F.U) ≈ Matrix(one(elty)*I, m, m)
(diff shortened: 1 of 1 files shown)


test/solver.jl
```diff
@@ -179,8 +179,8 @@ k = 1
         h_U            = collect(d_U)
         h_Vt           = collect(d_Vt)
         F              = svd(A, full=true)
-        @test abs.(h_U'h_U) ≈ Matrix(one(elty)*I, m, m)
-        @test abs.(h_U[:,1:n]'F.U[:,1:10]) ≈ Matrix(one(elty)*I, n, n)
+        @test abs.(h_U'F.U) ≈ Matrix(one(elty)*I, m, m)
```
andreasnoack: The old version was correct. The last `m-n` columns of `h_U` need not be orthogonal to the last `m_n` columns of `F.U`.

mbeltagy: Makes sense.

PR #178, opened 2018-10-31
Add optimized implementation of fill for types supported by memset.

maleadt:
Before: our broadcasting `fill!` is slower than a memory copy...

```
julia> @btime cuzeros(Float32,10,10)
  6.552 μs (47 allocations: 1.61 KiB)

julia> @btime CuArray(zeros(Float32,10,10))
  4.616 μs (18 allocations: 1.94 KiB)
```

After:

```
julia> @btime cuzeros(Float32,10,10)
  2.162 μs (15 allocations: 352 bytes)
```

Would be nice if we could dispatch on `haskey(MemsetTypes, sizeof(T))` instead of enumerating like now.

### DIFF
--- a/src/array.jl
+++ b/src/array.jl
@@ -148,6 +148,17 @@ cuones(T::Type, dims...) = fill!(CuArray{T}(undef, dims...), 1)
 cuzeros(dims...) = cuzeros(Float32, dims...)
 cuones(dims...) = cuones(Float32, dims...)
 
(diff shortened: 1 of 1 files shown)


maleadt:
Probably should implement this with a constructor instead of `cuzeros`, ref. https://github.com/JuliaLang/julia/issues/11557 ?

maleadt:
Let's merge this since it's a strict improvement.

Issue #176, opened 2018-10-30
Wrapping SVD can not be done at the moment due to restrictive struct definition in Base


mbeltagy:
I was trying to write a convenience function to SVD 
```julia
using CuArrays, CuArrays.CUSOLVER
import LinearAlgebra.svd
function svd(d_A::CuMatrix{T}) where T
    A=collect(d_A)
    d_U, d_S, d_Vt = CUSOLVER.gesvd!('A','A',d_A)
    SVD( d_U, d_S, d_Vt)
end
```
When I ran into a problem... The definition of the SVD struct from https://github.com/JuliaLang/julia/blob/master/stdlib/LinearAlgebra/src/svd.jl will not allow it. 

```julia
# Singular Value Decomposition
struct SVD{T,Tr,M<:AbstractArray{T}} <: Factorization{T}
    U::M
    S::Vector{Tr}
    Vt::M
    function SVD{T,Tr,M}(U, S, Vt) where {T,Tr,M<:AbstractArray{T}}
        @assert !has_offset_axes(U, S, Vt)
        new{T,Tr,M}(U, S, Vt)
    end
end
```

This might work if in the above definition is relaxed to
```
 S::AbstractVector{Tr}
```
It that something worth pursuing upstream in base?

andreasnoack:
Indeed the base version is too restrictive. However, generally it is not a good solution to have fields with abstract types since it operations can no longer be inferred. Instead, it could be made parametric similarly to `M`. For now, we should probably just have a `CuSVD` that mimics the `SVD` struct the same way that we have `CuQR`.

PR #183, opened 2018-11-03
Added a LinearAlgebra.svd! method and related struct CuSVD

mbeltagy:
Trying to resolve https://github.com/JuliaGPU/CuArrays.jl/issues/176

### DIFF
--- a/src/solver/highlevel.jl
+++ b/src/solver/highlevel.jl
@@ -61,3 +61,28 @@ function Base.show(io::IO, F::CuQR)
     println(io)
     show(io, F.R)
 end
+
+# Singular Value Decomposition
+
+struct CuSVD{T,Tr,M<:AbstractArray{T}} <: LinearAlgebra.Factorization{T}
+    U::M
+    S::CuVector{Tr}
+    Vt::M
+    function CuSVD{T,Tr,M}(U, S, Vt) where {T,Tr,M<:AbstractArray{T}}
+        new{T,Tr,M}(U, S, Vt)
+    end
+end
+CuSVD(U::AbstractArray{T}, S::CuVector{Tr}, Vt::AbstractArray{T}) where {T,Tr} = CuSVD{T,Tr,typeof(U)}(U, S, Vt)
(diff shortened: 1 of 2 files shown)


src/solver/highlevel.jl
```diff
@@ -61,3 +61,28 @@ function Base.show(io::IO, F::CuQR)
     println(io)
     show(io, F.R)
 end
+
+# Singular Value Decomposition
+
+struct CuSVD{T,Tr,M<:AbstractArray{T}} <: LinearAlgebra.Factorization{T}
+    U::M
```
andreasnoack: I think it's fine to make these `CuMatrix{T}` to simplify the type. It's unlikely that it'll be anything else for `CuiSVD`.

mbeltagy: That was my thinking initially, but noticed the `CuQR` and `CuQRPacked` are following the `AbstractMatrix` pattern. Perhaps we should change it there as well.

mbeltagy: Fixed in last commit.

maleadt:
CI looks good, let's merge. Thanks!

src/solver/highlevel.jl
andreasnoack: > Fixed in last commit.

It doesn't seems so. The `struct` hasn't changed.

mbeltagy: It has change in the commit below. :-)  
https://github.com/JuliaGPU/CuArrays.jl/pull/183/commits/22fe25094b51e96d928c568680de1472121e07a4

andreasnoack: I think we are talking past each other. Either that or I don't see the same diff as you. That commit doesn't change the struct.

mbeltagy: @andreasnoack 
You are absolutely right. Apologies. I missed that bit. 
Created a new pull request, since this one was merged already. 
https://github.com/JuliaGPU/CuArrays.jl/pull/187

PR #185, opened 2018-11-05
Try out Bors

maleadt:
As Bors mirrors commits on the repo itself, this should make it possible to build changes from forks.


maleadt:
bors try

PR #187, opened 2018-11-05
Missed a specialization for CuSVD CuMatrix... fixed now.

mbeltagy:
An addendum of missed code to https://github.com/JuliaGPU/CuArrays.jl/pull/183

### DIFF
--- a/src/solver/highlevel.jl
+++ b/src/solver/highlevel.jl
@@ -64,15 +64,11 @@ end
 
 # Singular Value Decomposition
 
-struct CuSVD{T,Tr,M<:CuMatrix{T}} <: LinearAlgebra.Factorization{T}
-    U::M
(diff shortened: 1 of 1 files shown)


src/solver/highlevel.jl
```diff
@@ -64,11 +64,11 @@ end
 
 # Singular Value Decomposition
 
-struct CuSVD{T,Tr,M<:AbstractArray{T}} <: LinearAlgebra.Factorization{T}
+struct CuSVD{T,Tr,M<:CuMatrix{T}} <: LinearAlgebra.Factorization{T}
```
andreasnoack: What I had in mind was
```julia
struct CuSVD{T,Tr} <: LinearAlgebra.Factorization{T}
    U::CuMatrix{T}
    S::CuVector{Tr}
    Vt::CuMatrix{T}
end
```
since then there would be only two type parameters.

PR #196, opened 2018-11-12
Inplace sparse multiplication

mohamed82008:
This PR implements `mul!` for sparse matrices on the GPU. It is my first time messing around with these stuff and the tests seem to fail for the whole package, not just what I added here, so please take your time reviewing this.


maleadt:
bors try
IIUC it should also be possible to remove the non-inplace versions and have Base allocate output containers and dispatch to these new methods.

src/sparse/highlevel.jl
```diff
@@ -19,6 +19,23 @@
...
+mul!(C::CuVector{T},adjA::Adjoint{<:Any, HermOrSym{T,CuSparseMatrix{T}}},B::CuVector{T}) where {T} = mv!('C',T(1),adjA.parent,B,T(0),C,'O')
+
+mul!(C::CuMatrix{T},A::CuSparseMatrix{T},B::CuMatrix{T}) where {T} = mm2!('N','N',T(1),A,B,T(0),C,'O')
+mul!(C::CuMatrix{T},A::CuSparseMatrix{T},transB::Transpose{<:Any, CuMatrix{T}})  where {T} = mm2!('N','T',T(1),A,transB.parent,T(0),C,'O')
+mul!(C::CuMatrix{T},transA::Transpose{<:Any, CuSparseMatrix{T}},B::CuMatrix{T})  where {T} = mm2!('T','N',T(1),transA.parent,B,T(0),C,'O')
+mul!(C::CuMatrix{T},transA::Transpose{<:Any, CuSparseMatrix{T}},transB::Transpose{<:Any, CuMatrix{T}}) where {T} = mm2!('T','T',T(1),transA.parent,transB.parent,T(0),C,'O')
+mul!(C::CuMatrix{T},adjA::Adjoint{<:Any, CuSparseMatrix{T}},B::CuMatrix{T})  where {T} = mm2!('C','N',T(1),adjA.parent,B,T(0),C,'O')
+
+mul!(C::CuMatrix{T},A::HermOrSym,B::CuMatrix) where {T} = mm!('N',T(1),A,B,T(0),C,'O')
+mul!(C::CuMatrix{T},transA::Transpose{<:Any, <:HermOrSym{<:Number, <:CuSparseMatrix}},B::CuMatrix) where {T} = mm!('T',T(1),transA.parent,B,T(0),C,'O')
```
maleadt: better use `parent(...)` here

src/sparse/highlevel.jl
```diff
@@ -19,6 +19,23 @@
 (*)(transA::Transpose{<:Any, <:HermOrSym{<:Number, <:CuSparseMatrix}},B::CuMatrix) = mm('T',transA.parent,B,'O')
 (*)(adjA::Adjoint{<:Any, <:HermOrSym{<:Number, <:CuSparseMatrix}},B::CuMatrix) = mm('C',adjA.parent, B,'O')
 
+mul!(C::CuVector{T},A::CuSparseMatrix,B::CuVector) where {T} = mv!('N',T(1),A,B,T(0),C,'O')
+mul!(C::CuVector{T},transA::Transpose{<:Any, CuSparseMatrix},B::CuVector) where {T} = mv!('T',T(1),transA.parent,B,T(0),C,'O')
```
maleadt: since T is unconstrained, `zero(T)` would be better here

src/sparse/highlevel.jl
```diff
@@ -19,6 +19,23 @@
...
+mul!(C::CuVector{T},adjA::Adjoint{<:Any, CuSparseMatrix},B::CuVector) where {T} = mv!('C',T(1),transA.parent,B,T(0),C,'O')
+mul!(C::CuVector{T},A::HermOrSym{T,CuSparseMatrix{T}},B::CuVector{T}) where T = mv!('N',T(1),A,B,T(0),C,'O')
+mul!(C::CuVector{T},transA::Transpose{<:Any, HermOrSym{T,CuSparseMatrix{T}}},B::CuVector{T}) where {T} = mv!('T',T(1),transA.parent,B,T(0),C,'O')
+mul!(C::CuVector{T},adjA::Adjoint{<:Any, HermOrSym{T,CuSparseMatrix{T}}},B::CuVector{T}) where {T} = mv!('C',T(1),adjA.parent,B,T(0),C,'O')
+
+mul!(C::CuMatrix{T},A::CuSparseMatrix{T},B::CuMatrix{T}) where {T} = mm2!('N','N',T(1),A,B,T(0),C,'O')
+mul!(C::CuMatrix{T},A::CuSparseMatrix{T},transB::Transpose{<:Any, CuMatrix{T}})  where {T} = mm2!('N','T',T(1),A,transB.parent,T(0),C,'O')
+mul!(C::CuMatrix{T},transA::Transpose{<:Any, CuSparseMatrix{T}},B::CuMatrix{T})  where {T} = mm2!('T','N',T(1),transA.parent,B,T(0),C,'O')
+mul!(C::CuMatrix{T},transA::Transpose{<:Any, CuSparseMatrix{T}},transB::Transpose{<:Any, CuMatrix{T}}) where {T} = mm2!('T','T',T(1),transA.parent,transB.parent,T(0),C,'O')
```
maleadt: `one(T)`

test/sparse.jl
```diff
@@ -1465,6 +1465,344 @@ end
...
+        A = sparse(rand(elty,m,k))
+        B = rand(elty,k,n)
+        C = rand(elty,m,n)
+        alpha = rand(elty)
+        beta = rand(elty)
+        @testset "csr" begin
+            d_B = CuArray(B)
+            d_C = CuArray(C)
+            d_A = CuSparseMatrixCSR(A)
+            CUSPARSE.mm!('N',alpha,d_A,d_B,beta,d_C,'O')
```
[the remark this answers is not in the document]
mohamed82008: IIUC. the other tests are for out of place versions of the functions, namely `mv`, `mm` and `mm2` for normal, adjoint, transpose, symmetric and Hermitian cases. So yes I did copy the tests from above, but I changed them to the in-place versions. I now notice that some of the previous tests which perhaps should be testing out of place functions are actually testing the inplace ones, e.g: https://github.com/JuliaGPU/CuArrays.jl/blob/5444129e3b5a730237c3a5110dfeaad53cbd3605/test/sparse.jl#L1078. However, this is only the case for symmetric and Hermitian `mm`. If this is not a bug, I will remove the equivalent tests from the `mm!` test set.

maleadt:
bors try

mohamed82008:
The new tests all pass on my machine now. Some old tests don't but that's for a different issue. It would be nice if bors gave this a second shot.

maleadt:
Great, thanks! @kshyatt, any final comments or is this good to go?

PR #206, opened 2018-11-19
Simplify package matching.

maleadt:


MikeInnes:
Just a thought, but perhaps we should just check in a manifest for this purpose? CI can then always use a known-good set of packages/git revisions as chosen by PR authors, and we can reset to releases just before tagging.

maleadt:
But then CI wouldn't detect issues caused by package upgrades. I also don't know how to efficiently generate a Manifest for eg. CuArrays; I'm developing those packages in a shared environment so my manifest contains many more packages, and I `dev` most packages so there's no tree hash.

vchuravy:
I liked having the ability to jointly push branches and test development that way.

maleadt:
> I liked having the ability to jointly push branches and test development that way.

As do I, but with the proposed Bors workflow the branch is always `trying` or `staging`...

vchuravy:
Ah ok. This makes sense then. So we are going full bors?

maleadt:
> So we are going full bors?

It sees like the only way to support CI for non-JuliaGPU members, which arguably is a more important feature than experimental branch matching.

maleadt:
I'm suspecting a library version error, so adding some code to report that.

maleadt:
bors try

maleadt:
Interesting. Had forgotten to run `ldconfig` after updating CUDA so it had picked up an incompatible mix of libraries.

maleadt:
Seems like I need to restart the process.
bors r+

maleadt:
Doc failure is CUDAapi not being tagged. Merging!

PR #209, opened 2018-11-20
Make bors monitor both branches

notriddle:
This allows you to have bors merge the branches in,
instead of only being able to use `try`.


maleadt:
Thanks for looking into this! I had actually added that change in https://github.com/JuliaGPU/CuArrays.jl/pull/206, but since that branch has some testing errors I'll go ahead and merge it here :slightly_smiling_face:

PR #213, opened 2018-11-24
tests for GPUArrays bug fix

GiggleLiu:
see
https://github.com/JuliaGPU/GPUArrays.jl/pull/176


maleadt:
This should only contain CuArray-specific tests (ie. the SubArray stuff).

PR #222, opened 2018-12-08
Forward-port NNLIB changes

vchuravy:
https://github.com/FluxML/NNlib.jl/pull/54 is dropping a unused parameter

PR #223, opened 2018-12-09
Changes in Conv API

tejank10:
The changes to the `conv` API of NNlib have been proposed in [#54](https://github.com/FluxML/NNlib.jl/pull/54). This PR makes the corresponding changes in order to make it compatible with CuArrays.

### DIFF
--- a/src/dnn/libcudnn.jl
+++ b/src/dnn/libcudnn.jl
@@ -296,7 +296,7 @@ function cudnnConvolutionBackwardData(alpha, wDesc, w, dyDesc, dy, convDesc, alg
                  workspace_size, beta, dxDesc, dx)
 end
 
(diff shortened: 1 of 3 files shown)


vchuravy:
In #222 I also had a compatibility layer to current NNLib. Not sure if that is really needed but it would allow us to merge this before the NNLib changes land.

vchuravy:
bors try

MikeInnes:
bors try

MikeInnes:
Note that this is an API change; I'll change the current version number so it's clear we need a new minor.

maleadt:
Doesn't this need a REQUIRE update?

MikeInnes:
As in, bumping the NNlib version? We do need to but NNlib isn't tagged yet. Don't know if there's any downside to just putting a future version in though, given that we do CI with master anyway.

maleadt:
> Don't know if there's any downside to just putting a future version in though, given that we do CI with master anyway.

Yeah, I typically do it early not to forget it when we end up tagging. But it's the same to me, I'm on `master` anyway.

PR #226, opened 2018-12-11
Allow unsafe free

MikeInnes:
This is a hack but I'm just putting it up for discussion (and so https://github.com/denizyuret/Knet.jl/pull/396 can use it).

Knet wants to be able to early-free pointers where it can. Right now this leads to an assertion error in the CuArrays finaliser. I'm not sure how best to handle this without just removing the assert entirely. @maleadt would be great to hear your thoughts.

### DIFF
--- a/src/array.jl
+++ b/src/array.jl
@@ -22,6 +22,7 @@ CuMatrix{T} = CuArray{T,2}
 CuVecOrMat{T} = Union{CuVector{T},CuMatrix{T}}
 
 function unsafe_free!(xs::CuArray)
+  Mem.refcounts[xs.buf] == 0 && return
   Mem.release(xs.buf) && dealloc(xs.buf, prod(xs.dims)*sizeof(eltype(xs)))
   return
 end


maleadt:
I assume performance of `finalize(obj)` is bad? An API like this would be fine, although this one right now is a little hacky of course. Maybe add a `# HACK` comment, LGMT otherwise.

MikeInnes:
I don't know what the semantics of `finalize` are; will it not be called by the GC then? If not then that could be a perfectly reasonable way to do this.

vchuravy:
In finalize you need to check if the object is still valid. As an example
one could pull the pointer (I think that's how we handle it in OpenCL).

maleadt:
> In finalize you need to check if the object is still valid.

That's not how I understand it. AFAIK `finalize` early-runs the finalizer, and it doesn't get to run again:

```julia
julia> mutable struct Foo
         x
       end

julia> a = Foo(1)
Foo(1)

julia> finalizer(obj->println(obj.x), a)
Foo(1)

julia> a = nothing

julia> GC.gc()
1

julia> b = Foo(2)
Foo(2)

julia> finalizer(obj->println(obj.x), b)
Foo(2)

julia> finalize(b)
2

julia> b = nothing

julia> GC.gc()
```

maleadt:
Upstream KNet branch uses `finalize` now.

PR #228, opened 2018-12-14
Wrap and test gesvdj!

kshyatt:
Not sure what's up with that broken test.


kshyatt:
bors try

test/solver.jl
```diff
@@ -189,6 +189,20 @@ k = 1
...
+    
+    @testset "gesvdj!" begin
+        A              = rand(elty,m,n)
+        d_A            = CuArray(A)
+        d_U, d_S, d_V  = CUSOLVER.gesvdj!('V',1,d_A)
+        h_S            = collect(d_S)
+        h_U            = collect(d_U)
+        h_V            = collect(d_V)
+        F              = svd(A, full=true)
+        @test_broken abs.(h_U'h_U) ≈ Matrix(one(elty)*I, m, m) atol=1e-6
```
andreasnoack: Why is this broken?

andreasnoack: We should actually make sure to test both the tall and wide case for both SVDs. It's so easy to make small errors with the dimensions.

kshyatt: Because the off-diagonal elements are sometimes pretty large

kshyatt: Just to check, you mean prepare A with dimensions `(m, n)` and `(n, m)` where `m > n`, right?

andreasnoack: Yes, exactly.

kshyatt:
@andreasnoack what do you think, good to go?

test/solver.jl
```diff
@@ -189,6 +189,28 @@ k = 1
...
+            d_U, d_S, d_V  = CUSOLVER.gesvdj!('V',1,d_A)
+            h_S            = collect(d_S)
+            h_U            = collect(d_U)
+            h_V            = collect(d_V)
+            F              = svd(A, full=true)
+            s              = min(a, b)
+            t              = max(a, b)
+            @test size(h_V) == (b, s) 
+            @test size(h_U) == (a, s) 
+            tol = elty == Float32 || elty == ComplexF32 ? 1e-4 : 1e-8
```
andreasnoack: It's crazy that the basis is so skewed. Wondering if this is because of the Jacobi algorithm or if it's their implementation. Anyway, I think the PR is fine.

kshyatt: It could be the optional parameters I'm giving it.

PR #229, opened 2018-12-14
More tests for generalized eigen methods

kshyatt:
Should also test that the no vector version works and I totally forgot about the symmetric generalized eigenproblem!


kshyatt:
bors try

PR #233, opened 2018-12-20
Add an introductory tutorial

timholy:
https://htmlpreview.github.io/?https://github.com/JuliaGPU/CuArrays.jl/blob/teh/tutorial/tutorials/build/intro.html

I don't know if we want to commit the html or build it dynamically,  but here's at least a start.
Comments/corrections appreciated.


maleadt:
Nice! If dumped them somewhere in `public/` during CI build (.gitlab.yml) they'll be hosted at https://juliagpu.gitlab.io/CuArrays.jl/. Would probably be better to not ship the HTML and instead add a Project.toml to `tutorials/` like `docs/` is structured nowadays?

timholy:
OK, try this. With this version, I am struggling and failing to get the image to embed no matter how I specify the path. I've tried basically every variant I can think of, short of a permalink to the hosted png file once this gets merged. If anyone knows how to fix this...

Two other items to check:
- regarding the image: is "borrowing" this OK or should I redraw something from scratch?
- checks on factual content appreciated. In particular, is my statement in the error-handling section right, that typically the only way to recover from errors is to restart Julia? (Presumably https://github.com/JuliaGPU/CUDAnative.jl/pull/303 might change that?)

Ken-B:
Your original link doesn't work for me anymore, [this one does](http://htmlpreview.github.io/?https://github.com/JuliaGPU/CuArrays.jl/blob/a689c0986ba7fe7442878a8155838ec3f4b8afb0/tutorials/build/intro.html).

maleadt:
Rebased and changed the infrastructure a little (I think it broke the docs).
bors try

maleadt:
Those are some weird crashes... But anyhow, forgot to add a file, so let's try again.
bors try

maleadt:
Another typo. Should be fine now
bors try

maleadt:
PoC: a `script` function to execute code outside of the Weave session but include its output. Demonstrated here to execute `nvprof`, but could also be useful for eg. encoding the illegal access situation where the GPU dies.
bors try

maleadt:
Forgot some more deps.
bors try

maleadt:
Reworked the tutorial a little.
bors try

timholy:
Thanks for all that polish! Didn't mean to assign you a bunch of work.

For your PR "improve the image," did you actually edit it? If we're editing it, we could change the indexing from 0 to 1. I'm willing to do this if you think it's worth it.

maleadt:
> For your PR "improve the image," did you actually edit it? If we're editing it, we could change the indexing from 0 to 1. I'm willing to do this if you think it's worth it.

I just quickly edited out the red squiggly lines from the spell checker, they bothered me too much...

xiaodaigh:
Got a question, so the number of threads and number of blocks is chosen by the user? I gather, from the tutorial that it isn't something related to the hardware? Presumably, the `broadcast` function would need to choose the number of threads and the number of blocks for the user automatically. How is that choosen?

xiaodaigh:
In the addition GPU kernel example, the vectors are called x and y and the thread_ids and block ids use the `x` coordinates as well. I think it's better to rename the vectors to `a` and `b` or some other variable name that don't clash with the indexing coordinate, to make absolutely certain that they are different to avoid any possibility of confusion.

maleadt:
> the number of threads and number of blocks is chosen by the user? I gather, from the tutorial that it isn't something related to the hardware?s 

It's an introductory tutorial so doesn't explain everything. The user chooses, yes, but there's hardware limits and performance characteristics to take into account. Broadcast picks pretty naively, https://github.com/JuliaGPU/CuArrays.jl/blob/249c6e96c450f4e367f2fc4c48999ca7bb2d2069/src/utils.jl#L3-L6, other implementations already do something more sophisticated https://github.com/JuliaGPU/CuArrays.jl/blob/249c6e96c450f4e367f2fc4c48999ca7bb2d2069/src/mapreduce.jl#L70-L83

timholy:
FYI, I'm fine with merging this whenever you feel ready, if all the functionality and description is how you want it. (Though it doesn't seem to render now, can't take time to investigate unfortunately.)

maleadt:
bors try

maleadt:
Ah, some issues with CI. Looks like I'll be spamming again.
bors try

maleadt:
bors try

maleadt:
bors try

maleadt:
Note: will only be live once CI finishes in on the master branch.

PR #234, opened 2018-12-20
Small improvements to blas wrappers and tests

kshyatt:


kshyatt:
bors try

kshyatt:
! Was it something in this PR that caused the `map reduce` fail? This passed locally for me...

PR #239, opened 2018-12-30
Fix bug in gemm strided batched

avik-pal:
MWE:
```julia
using CuArrays
a = cu(rand(4, 3, 2))
b = cu(rand(3, 4, 2))
CuArrays.CUBLAS.gemm_strided_batched('T', 'T', a, b)
```
From my understanding this should give a `3 x 3 x 2` array. Instead we get a `Dimension Mismatch`.


vchuravy:
Mind adding a test for this?

avik-pal:
Sure will do.

PR #243, opened 2019-01-03
add a setting for the CUBLAS math mode to enable using tensor gemm

### LINKED ISSUE #242: Feature request: Allow calling cublasSetMathMode
On newer graphics card with tensor cores, this could improve performance for some inputs (https://devblogs.nvidia.com/programming-tensor-cores-cuda-9/).

KristofferC:
Fixes #242 

Not sure where to put this / where to document it. Should probably add a test as well.

As an example of performance difference using `m=n=k=2048` with `Float32` on a V-100:

```
            Type  Time(%)      Time     Calls       Avg       Min       Max  Name
 GPU activities:   81.70%  1.5610ms         1  1.5610ms  1.5610ms  1.5610ms  volta_sgemm_128x64_nn
                   18.06%  345.02us         1  345.02us  345.02us  345.02us  volta_s884gemm_128x128_ldg8_f2f_nn
```

### DIFF
--- a/src/blas/CUBLAS.jl
+++ b/src/blas/CUBLAS.jl
@@ -28,13 +28,6 @@ function handle()
     return _handle[]
 end
 
-@enum CUBLASMathMode::Cint begin
-   CUBLAS_DEFAULT_MATH = 0
-   CUBLAS_TENSOR_OP_MATH = 1
-end
-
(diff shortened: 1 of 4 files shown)


KristofferC:
But do we want to do that? It changes the results slightly so perhaps it is better as an opt-in feature?

maleadt:
Mm right, hadn't properly read the docs. Maybe tie it into `--math-mode` by inspecting `JLOptions().fast_math`, or is that too much of a pun?

PR #245, opened 2019-01-03
Improve at-sync performance.

maleadt:
From https://discourse.julialang.org/t/most-efficient-way-of-waiting-for-gpu-results/19193
bors try

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -1,5 +1,11 @@
 using Documenter, CuArrays
 
+using Pkg
+if haskey(ENV, "GITLAB_CI")
+    Pkg.add([PackageSpec(name = x; rev = "master")
+             for x in ["CUDAapi", "GPUArrays", "CUDAnative", "NNlib", "CUDAdrv"]])
+end
+
 makedocs(
     modules = [CuArrays],
     format = Documenter.HTML(prettyurls = get(ENV, "CI", nothing) == "true"),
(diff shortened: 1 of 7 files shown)


maleadt:
bors try

maleadt:
Hmm, not sure why the master version of CUDAdrv isn't used.
bors try

maleadt:
bors try

maleadt:
Finally.
bors r+

docs/make.jl
```diff
@@ -1,5 +1,10 @@
 using Documenter, CuArrays
 
+using Pkg
+if haskey(ENV, "GITLAB_CI")
+    Pkg.develop.(["CUDAapi", "GPUArrays", "CUDAnative", "NNlib", "CUDAdrv"])
```
KristofferC: I would add these at the master branch instead (and also fuse it to a single call to `Pkg.add`)

```jl
Pkg.add([PackageSpec(name = x; rev = "master") for x in ["CUDAapi", "GPUArrays", "CUDAnative", "NNlib", "CUDAdrv"]])```

File: test/runtests.jl [hunk 2]
```diff
@@ -1,5 +1,12 @@
 using Test
 
+# development often happens in lockstep with other packages,
+# so check-out the master branch of those packages.
+using Pkg
+if haskey(ENV, "GITLAB_CI")
```
KristofferC: Same here.

docs/make.jl
maleadt: Ah yeah that's probably better, I had changed it to `develop` in an attempt to test the issue (where I did the `Pkg.add` as part of `Pkg.build`).

Issue #253, opened 2019-01-14
broadcast custom function break intrinsics rewrite and crash CUDA


chengchingwen:
Similar to #154 but some function also cause error and crash cuda. 
I'm writing custom activation functions like `NNlib.jl` do but encounter some error while implement some functions. Here is a simplest example to reproduce the error. 
I'm using `CuArrays v0.8.1` 

```julia
using CuArrays
f(x) = exp(x)
g(x) = tanh(x)
h(x) = log(x)

julia> f.(cu(randn(10,10)))
┌ Warning: calls to Base intrinsics might be GPU incompatible
│   exception =
│    You called exp(x::T) where T<:Union{Float32, Float64} in Base.Math at special/exp.jl:75, maybe you intended to call exp(x::Float32) in CUDAnative at /home/peter/.julia/packages/CUDAnative/AGfq2/src/device/libdevice.jl:90 instead?
│    Stacktrace:
│     [1] exp at special/exp.jl:75
│     [2] #25 at /home/peter/.julia/packages/GPUArrays/hzyWn/src/broadcast.jl:58
└ @ CUDAnative ~/.julia/packages/CUDAnative/AGfq2/src/compiler/irgen.jl:111
10×10 CuArray{Float32,2}:
 4.55366   0.218534  16.5519    1.43041   0.650593  4.68102   0.470166  0.917978  1.5589    0.638304
 1.3706    4.57008    0.26529   2.33283   1.02681   2.93923   0.291703  1.704     2.23857   0.102735
 0.328412  0.345686   0.331952  0.163646  0.742078  0.471301  1.62113   0.511803  0.357865  0.435097
 0.91065   1.91572    1.0373    3.78657   0.752613  0.326961  0.491546  0.643741  0.553853  4.89362
 2.1674    0.226056   0.410075  0.503229  1.21052   0.813538  9.2301    1.08758   0.183459  3.50125
 1.23666   4.43529    1.60278   2.47632   0.792438  1.02627   2.94931   0.767341  2.21165   0.168986
 0.734377  5.69426    0.783176  3.77377   0.109799  0.689816  0.385528  0.688056  1.83929   0.504368
 0.715148  1.15716    1.48569   0.320102  0.441299  0.519798  0.4096    1.65278   1.46279   0.927808
 8.17749   1.33664    0.089473  0.164976  1.10891   0.323992  0.303008  1.14196   3.41736   0.241758
[...]

PR #254, opened 2019-01-15
broadcast rewrite with custom functinos

chengchingwen:
As mentioned in #253, support `@cufunc` to be used with some custom functions that had already registered.

### DIFF
--- a/src/broadcast.jl
+++ b/src/broadcast.jl
@@ -39,17 +39,23 @@ end
 
 using MacroTools
 
+const _cufuncs = copy(libdevice)
+cufuncs() = (global _cufuncs; _cufuncs)
+
 function replace_device(ex)
+  global _cufuncs
   MacroTools.postwalk(ex) do x
-    x in libdevice ? :(CUDAnative.$x) : x
+    x in _cufuncs ? :(CuArrays.cufunc($x)) : x
   end
 end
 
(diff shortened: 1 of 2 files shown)


test/base.jl
```diff
@@ -77,6 +77,24 @@ end
   @test Array(Whatever{Int}.(CuArray([1]))) == Whatever{Int}.([1])
 end
 
+@testset "Cufunc" begin
+  import CUDAnative
```
maleadt: why is this needed?

maleadt:
bors try

test/base.jl
chengchingwen: you are right, I forget that I change the implementation.

src/broadcast.jl
```diff
@@ -39,9 +39,12 @@ end
 
 using MacroTools
 
+const _cufuncs = copy(libdevice)
+cufuncs() = (global _cufuncs; _cufuncs)
+
 function replace_device(ex)
   MacroTools.postwalk(ex) do x
-    x in libdevice ? :(CUDAnative.$x) : x
+    x in cufuncs() ? :(CuArrays.cufunc($x)) : x
```
MikeInnes: You could replace this with `x in _cufuncs` and get rid of the indirection here

PR #258, opened 2019-01-18
Fix svd wrappers and high level interface

andreasnoack:
...such that `svd` and `svdvals` now work. Also, enable an option for using the Jacobi algorithm and make it the default since it is more efficient on GPUs.


andreasnoack:
bors try

andreasnoack:
bors try

andreasnoack:
bors try

andreasnoack:
I should mention that I've adjusted the `CuSVD` struct such that it can store both `Adjoint(Vt)` from the QR based solver and `V` from the Jacobi based solver. To avoid that the type instability becomes a problem, the algorithm is selected with a positional argument instead of a keyword argument. Alternatively, we'd have to use types.

PR #263, opened 2019-01-21
add cudnnSetConvolutionMathType

KristofferC:
Not sure what the best way to add a test is. Something stupid like

```jl
julia> convDesc = CuArrays.CUDNN.ConvDesc(Float32, 2, 0, 1, 1, 0);;

julia> CuArrays.CUDNN.cudnnSetConvolutionMathType(convDesc, CuArrays.CUDNN.CUDNN_TENSOR_OP_MATH_ALLOW_CONVERSION);

julia> @test_throws CuArrays.CUDNN.CUDNNError CuArrays.CUDNN.cudnnSetConvolutionMathType(convDesc, 3)
Test Passed
      Thrown: CuArrays.CUDNN.CUDNNError
```

?

### DIFF
--- a/src/dnn/libcudnn.jl
+++ b/src/dnn/libcudnn.jl
@@ -119,6 +119,13 @@ function cudnnDestroyConvolutionDescriptor(convDesc)
                  convDesc)
 end
 
+function cudnnSetConvolutionMathType(convDesc, mathType)
+    @check ccall((:cudnnSetConvolutionMathType,libcudnn),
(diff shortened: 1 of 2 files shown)


vchuravy:
Teah that looks like an ok test

maleadt:
Eh, we don't really test the internal `ccall`s unless there's a higher-level wrapper for it. Let's just merge this as is and add a proper wrapper/test when we need it .

PR #271, opened 2019-02-04
Adapt to the new CUDAdrv.CuPtr pointer type

maleadt:
See https://github.com/JuliaGPU/CUDAdrv.jl/pull/125

I'd appreciate some help on this; it's a pretty mechanical change: run the test-suite, spot conversion errors, have a look at the underlying `ccall` and change `Ptr` arguments to `CuDevicePtr` arguments when if they pass an `CuArray`.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,6 +1,6 @@
 julia 1.0
-CUDAnative 1.0.0
-CUDAdrv 1.0.1
+CUDAnative 1.1
+CUDAdrv 1.1
 CUDAapi 0.5.3
 NNlib 0.4.3
 GPUArrays 0.5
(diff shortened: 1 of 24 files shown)


maleadt:
OK, this wasn't fun. All but one test pass now.

maleadt:
I think this is starting to look good. Unless there's objections, I'll go ahead and merge this later today.

@MikeInnes this of course needs fixing of the cudnn/curnn wrappers in Flux.
Maybe now is a good time to move them into CuArrays.

Also, this generally has been a pretty fragile change where I had to peer the wrappers next to the NVIDIA docs and inspect every `Ptr` in every `ccall`. Needless to say, this is pretty error-prone. Testing has flushed out many errors, but our coverage is bad so expect fallout from this change.

MikeInnes:
Ok, I'll look at copying the Flux wrappers over soon.

maleadt:
bors try

Issue #273, opened 2019-02-05
gc(true) can be prohibitly slow to call


KristofferC:
I was profiling why a resnet model was running extremely slow on Flux.

Sprinkling some sections using https://github.com/KristofferC/TimerOutputs.jl and training the model a little bit I got:

```
 ───────────────────────────────────────────────────────────────────────
                                Time                   Allocations
                        ──────────────────────   ───────────────────────
    Tot / % measured:        82.6s / 62.9%           9.09GiB / 3.32%

 Section        ncalls     time   %tot     avg     alloc   %tot      avg
 ───────────────────────────────────────────────────────────────────────
 gc true           468    47.8s  92.1%   102ms   4.88MiB  1.58%  10.7KiB
 crossentropy       32    1.91s  3.69%  59.8ms    159MiB  51.5%  4.98MiB
 conv            1.70k    1.61s  3.10%   948μs   85.6MiB  27.7%  51.7KiB
 dense              32    488ms  0.94%  15.2ms   59.1MiB  19.1%  1.85MiB
 gc false           32   81.0ms  0.16%  2.53ms    645KiB  0.20%  20.2KiB
 ───────────────────────────────────────────────────────────────────────
```

The `gc true` section refers to only this line:

https://github.com/JuliaGPU/CuArrays.jl/blob/61e25a2d239da77a5e8f3dc9746f9f62cd9e1380/src/memory.jl#L256

It seems this line is being called too often compared to how expensive a `gc(true)` call is.

Issue #270, opened 2019-02-04
Eagerly put objects back in the pool


maleadt:
Basically, something like https://github.com/JuliaGPU/CuArrays.jl/pull/226, but more careful and with actual uses in CuArrays.jl. Some backstory.

We now depend on the Julia GC to collect `CuArray` objects before they end up in the pool again. This is not optimal, and compounded by the fact that the Julia GC doens't know about the GPU's memory pressure. As a result, we kind-of rely on hitting a GPU OOM before collecting dead arrays and putting them back into the pool: https://github.com/JuliaGPU/CuArrays.jl/blob/8525a9de7a8da922b0fc4be5783c5bdc5fb8a7a6/src/memory.jl#L254-L256

On the other hand, we often _know_ that arrays are unused and could be put back into the pool again. For example: https://github.com/JuliaGPU/CuArrays.jl/blob/0384d2530f64d808385bc67b8c1cf9f6b73e1c9a/src/solver/dense.jl#L126-L138

One option is to call `finalize`, but that is a pretty slow call and shouldn't be used in performance critical code. Instead, we should have a call to eagerly "free" objects (from the perspective of CuArrays.jl) and put them back into the pool. For example, the `geqrf!` method above would then free its `buffer` before returning. To prevent use of early-freed arrays, we should probably wipe the memory buffer handle (eg. set to NULL or some other sentinel value). Care should be taken to deal with ownership (not support this call when we don't own the `CuArray`? make sure the CUDAdrv refcount is 1?).

maleadt:
cc @SuperFluffy -- would be good to have a use-case or benchmark.

KristofferC:
The common way of doing this is (of course) with the `do` syntax, but this will likely make the code a lot uglier and more "indent"-y.

maleadt:
Implemented both.

PR #275, opened 2019-02-07
Support for early-freeing GPU memory

maleadt:
Implement https://github.com/JuliaGPU/CuArrays.jl/issues/270

There aren't too many opportunities for this within CuArrays, so it will probably not help too much towards https://github.com/JuliaGPU/CuArrays.jl/issues/273. Probably within Flux there might be more, but then we'd need to have a standardized method that can be called on different types of arrays.


maleadt:
bors try

PR #280, opened 2019-02-11
Add more convenient tracing functionality.

maleadt:
I'm not really happy with this implementation but it might be of some use to people. Calling `gc(true)` that often seems to mess with the Julia GC though, I've strangely seen OOM's happen because of it. And for https://github.com/KristofferC/resnet it generates so much output that it isn't really useful. Much of the wall-of-output seemed to trace to `Flux.back!` though so that might be of some interest.

bors try

### DIFF
--- a/src/memory.jl
+++ b/src/memory.jl
@@ -90,8 +90,10 @@ Base.copy(stats::PoolStats) =
   PoolStats((getfield(stats, field) for field in fieldnames(PoolStats))...)
 
 # allocation traces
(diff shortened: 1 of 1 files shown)


MikeInnes:
Yeah, Flux's backtraces are horrible since we effectively evaluate the whole program using a recursive interpreter. Potentially we could build fancier tooling to manipulate the backtraces there, but I'm also hoping we can just switch to Zygote, which has really good debug info.

PR #281, opened 2019-02-12
Value function iteration on the gpu tutorial

floswald:
hi there,
I really enjoyed the tutorial on starting out with GPU and I thought I'd give it a go for a very specific use case. value function iteration is a very common algorithm in macro economics, and people are interested in making it more efficient. I extended your existing tutorial in that direction. it's compiled here

https://floswald.github.io/html/vfi.html

I don't know whether you want to create a space for things like this somewhere, I'm unsure it could belong to your manual; anyway, I post this here, maybe only to get some feedback on it.

thanks!


maleadt:
On the one hand it's good to have more examples, especially commented like this, but this seems very specific for a tutorial. Maybe we should add a separate `examples` section where documented code like this would fit in. The example is also pretty heavy on kernel programming, so I don't know whether it would better fit in CUDAnative.jl or CuArrays.jl. Thoughts, @timholy @MikeInnes?

MikeInnes:
I think it's fine to consolidate our docs into CuArrays. It's easy to find then, and there's not much reason to look for it unless you are interested in doing something more advanced. I think we could probably wait to split out `examples` if/when we have a problem with too many tutorials, but don't have a strong opinion.

PR #282, opened 2019-02-12
Fix ambiguity in _reshape when dims is Tuple{Int}

ChrisRackauckas:
Since it's just the identity, the same array is just returned. Handles https://github.com/JuliaGPU/CuArrays.jl/issues/161 , https://github.com/JuliaDiffEq/DiffEqFlux.jl/issues/21 .

@vchuravy


maleadt:
@kshyatt this is probably what you ran into recently?
bors try

test/base.jl
```diff
@@ -167,6 +167,16 @@ end
   end
 end
 
+@testset "Reshape" begin
+  A = [1,2,3,4;5,6,7,8]
```
KristofferC: ?

PR #285, opened 2019-02-19
Add simple accumulate!

dpsanders:
Adds a simple implementation of `accumulate!`.
This unfortunately uses only global memory.


maleadt:
Nice! Fixes https://github.com/JuliaGPU/CuArrays.jl/pull/285

test/accumulate.jl
```diff
@@ -0,0 +1,12 @@
+using CuArrays
+
+v = [1, 2, 3]
+v2 = CuArray(v)
+
+@testset "accumulate! tests"
```
maleadt: maybe also directly test `accumulate`/`accumulate!`
we also typically use `testf` to compare against the `Array` output instead of hardcoding results.

maleadt:
How did you test this functionality? It doesn't seem like it executes at all, using none of the (original) test cases.

dpsanders:
Sorry about this, let me try to see what's going on.

maleadt:
It looks like the kernels aren't executed at all, and this is falling back to the scalar Base implementation. Probably needs to implement `Base._accumulate!` instead (or one of the other `_accumulate` functions).

dpsanders:
OK I think this is fixed now. I had changed `T.(v)` to `convert.(T, v)` without testing that change, and that seems to have caused the issue.

What's the correct way to convert a vector to a `Vector{T}`?

dpsanders:
Should I / can I run `bors` to run the tests?

dpsanders:
Suggested on Slack by @tlienart
```
vin = convert(CuVector{T}, v)  # possibly no-op
vin !== v || (vin = copy(v))
```

maleadt:
> Should I / can I run `bors` to run the tests?

Yeah go ahead, it is supposed to work by just saying:
bors try

( it might need a flag to allow non-JuliaGPU members to do so )

maleadt:
That said, I think that doesn't fix everything though. Not all `Base.accumulate` dispatch to the `accumulate!` method you've implemented here:

```
julia> using CuArrays
[ Info: Recompiling stale cache file /home/tbesard/Julia/depot/compiled/v1.1/CuArrays/7YFE0.ji for CuArrays [3a865a2d-5b23-5a0f-bc46-62713ec82fae]

julia> CuArrays.allowscalar(false)
false

julia> accumulate(+, CuArray([1]))
ERROR: scalar getindex is disallowed
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] assertscalar at /home/tbesard/Julia/GPUArrays/src/indexing.jl:8 [inlined]
 [3] getindex(::CuArray{Int64,1}, ::Int64) at /home/tbesard/Julia/GPUArrays/src/indexing.jl:44
 [4] first at ./abstractarray.jl:270 [inlined]
 [5] _accumulate!(::Function, ::CuArray{Int64,1}, ::CuArray{Int64,1}, ::Nothing, ::Nothing) at ./accumulate.jl:309
 [6] #accumulate!#586 at ./accumulate.jl:295 [inlined]
 [7] (::getfield(Base, Symbol("#kw##accumulate!")))(::NamedTuple{(:dims,),Tuple{Nothing}}, ::typeof(accumulate!), ::Function, ::CuArray{Int64,1}, ::CuArray{Int64,1}) at ./none:0
 [8] #accumulate#585(::Nothing, ::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::Function, ::Function, ::CuArray{Int64,1}) at ./accumulate.jl:247
 [9] accumulate(::Function, ::CuArray{Int64,1}) at ./accumulate.jl:239
 [10] top-level scope at none:0
```

dpsanders:
Ouch. I only needed `cumsum` so just implemented the relevant method of `accumulate!`.
Doing the general case seems pretty complicated.

src/accumulate.jl
```diff
@@ -0,0 +1,38 @@
...
+
+        vin, vout = vout, vin
+        Δ *= 2
+    end
+
+    return vin
+end
+
+
+function _partial_accumulate!(op, vout, vin, Δ)
```
KristofferC: `op` is unused here.

dpsanders: Thanks for finding that bug in the code!

dpsanders:
So apparently even `cumprod` is implemented differently from `cumsum` and does not work :/
I will try to do more general functions tomorrow or over the weekend.

MikeInnes:
We could potentially merged this with `op::typeof(+)` so it just catches the `cumsum` case for now. Just an option if it's a pain to do the more general thing.

dpsanders:
I think I now have all vector cases working (without specifying an initial value).

The matrix cases should be possible by calling the vector version repeatedly.

maleadt:
OK. feel free to comment-out the matrix tests and kick of CI. It's a strict improvement already, so it's fine if we only support vector without init for now.

maleadt:
Thanks!

dpsanders:
Thanks for guiding me through and for the feedback!

PR #289, opened 2019-02-20
cu(xs) respects DataType of xs.

ali-ramadhan:
I feel like having `cu(xs)` always return a `CuArray{Float32}` defied my expectations of what the function should do. If `xs` is an `Array{Float64}` or `Array{Complex{Float32}}` then the output of `cu(xs)` should reflect that.

This is especially helpful for doing mixed-precision GPGPU programming or for generically switching between `Float32` and `Float64`.

### DIFF
--- a/src/array.jl
+++ b/src/array.jl
@@ -205,7 +205,7 @@ end
 
 ## utilities
 
-cu(xs) = adapt(CuArray{Float32}, xs)
+cu(xs) = adapt(CuArray{eltype(xs)}, xs)
 Base.getindex(::typeof(cu), xs...) = CuArray([xs...])
 
 cuzeros(T::Type, dims...) = fill!(CuArray{T}(undef, dims...), 0)


maleadt:
That has been a conscious design decision by @MikeInnes, which I don't like either, but can't just be switched right now because lots of code relies on that behavior.

Why not just use the `CuArray` constructor?

ali-ramadhan:
Ah I see, fair enough. I am switching to using the constructor but was puzzling over why my Float64 simulations looked like Float32 simulations until I noticed cu(...) always returns Float32 arrays, which was not what I expected.

What if there was another function, e.g. (horrible name) `cu_preserve_datatype(xs)`?

maleadt:
> What if there was another function, e.g. (horrible name) `cu_preserve_datatype(xs)`?

Just call `adapt(CuArray, ...)` then, not sure it's worth another public API.

ali-ramadhan:
Ah that should work nicely. Thank you!

PR #294, opened 2019-02-25
Initial work towards cublasXt wrappers

kshyatt:
Something is going wrong after the tests run, all tests after the `cublasXt` ones fail due to the library not being initialized and I'm having a hell of a time figuring out why.

### DIFF
--- a/src/blas/CUBLAS.jl
+++ b/src/blas/CUBLAS.jl
@@ -43,9 +43,13 @@ function xt_handle()
         _xt_handle[] = get!(_xt_handles, active_context[]) do
             context = active_context[]
             dev = CUDAdrv.device(context)
-            xt_handle = cublasXtCreate(nDevices=1, deviceId=Cint[dev.handle], blockDim=Cint(64))
-            atexit(()->CUDAdrv.isvalid(context) && cublasXtDestroy(xt_handle) && setindex!(active_context, CUDAdrv.CuCurrentContext()))
-            xt_handle
+            handle = cublasXtCreate()
+            cublasXtDeviceSelect(handle, 1, [dev.handle])
+            #cublasXtSetBlockDim(handle, 64)
+
+            atexit(()->CUDAdrv.isvalid(context) && cublasXtDestroy(handle))
+
+            handle
         end
     end
     return _xt_handle[]
--- a/src/blas/CUBLAS.jl
+++ b/src/blas/CUBLAS.jl
@@ -1,11 +1,10 @@
 module CUBLAS
 
-import CUDAdrv: CUDAdrv, CuContext, CuStream_t, CuPtr, PtrOrCuPtr, CU_NULL
+import CUDAdrv: CUDAdrv, CuContext, CuStream_t, CuPtr, PtrOrCuPtr, CU_NULL, devices
 import CUDAapi
 
 using ..CuArrays
 using ..CuArrays: libcublas, active_context, unsafe_free!
-
 using LinearAlgebra
 
 include("libcublas_types.jl")
@@ -43,12 +42,11 @@ function xt_handle()
         _xt_handle[] = get!(_xt_handles, active_context[]) do
             context = active_context[]
             dev = CUDAdrv.device(context)
+            devs = map(x-> x.handle, CUDAdrv.devices()) 
             handle = cublasXtCreate()
-            cublasXtDeviceSelect(handle, 1, [dev.handle])
+            cublasXtDeviceSelect(handle, length(devs), devs)
             #cublasXtSetBlockDim(handle, 64)
-
             atexit(()->CUDAdrv.isvalid(context) && cublasXtDestroy(handle))
-
             handle
         end
     end
--- a/src/blas/CUBLAS.jl
+++ b/src/blas/CUBLAS.jl
@@ -43,8 +43,8 @@ function xt_handle()
         _xt_handle[] = get!(_xt_handles, active_context[]) do
             context = active_context[]
             dev = CUDAdrv.device(context)
-            xt_handle = cublasXtCreate(nDevices=1, deviceId=Cint[dev.handle]; blockDim=Cint(64))
-            atexit(()->CUDAdrv.isvalid(context) && cublasXtDestroy(xt_handle))
+            xt_handle = cublasXtCreate(nDevices=1, deviceId=Cint[dev.handle], blockDim=Cint(64))
+            atexit(()->CUDAdrv.isvalid(context) && cublasXtDestroy(xt_handle) && setindex!(active_context, CUDAdrv.CuCurrentContext()))
             xt_handle
         end
     end
--- a/src/blas/CUBLAS.jl
+++ b/src/blas/CUBLAS.jl
@@ -41,11 +41,9 @@ function xt_handle()
         @assert isassigned(active_context) # some other call should have initialized CUDA
         _xt_handle[] = get!(_xt_handles, active_context[]) do
             context = active_context[]
-            dev = CUDAdrv.device(context)
-            devs = map(x-> x.handle, CUDAdrv.devices()) 
             handle = cublasXtCreate()
+            devs = convert.(Cint, CUDAdrv.devices())
             cublasXtDeviceSelect(handle, length(devs), devs)
-            #cublasXtSetBlockDim(handle, 64)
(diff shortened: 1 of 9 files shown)


kshyatt:
bors try

kshyatt:
OK, so reordering the tests "worked" (and I fixed a broken one in CUSOLVER). Something about cublasXt and CUSOLVER makes them not want to play nicely.

kshyatt:
bors try

maleadt:
> OK, so reordering the tests "worked" (and I fixed a broken one in CUSOLVER). Something about cublasXt and CUSOLVER makes them not want to play nicely.

Well that is disconcerting. Could you reproduce locally? You mentioned something about creating a new context on Slack, could you elaborate?

kshyatt:
Yes, I can reproduce locally. If the CUBLAS tests run last everyone is happy but they seem to corrupt something for all later tests, and only when the `Xt` tests are run (thus, only when `cublasXtCreate` is called...).

kshyatt:
OK, updated again ... I think something is wrong with the finalizer for `xt` which would explain why everything worked when the BLAS tests run last. All the test failures seem to be in CUBLAS calls post the `xt` tests BEFORE I tried to reset the context here, it's almost all https://docs.nvidia.com/cuda/cusolver/index.html#cuSolverSPstatus INTERNAL_ERROR aka `async_timeout`. So I think there must be something funny happening to the primary context.

maleadt:
Did some debugging, looks like the cublasXt initialization mutates the active context:

```
CUDAdrv.CuCurrentContext() = CUDAdrv.CuContext(Ptr{Nothing} @0x00000000024bd030, true, true)
(CuArrays.CUBLAS).cublasXtDeviceSelect(handle, 1, Cint[0]) = 0x00000000
CUDAdrv.CuCurrentContext() = CUDAdrv.CuContext(Ptr{Nothing} @0x0000000001f86f30, false, true)
```

This of course messes with everything.

maleadt:
This is annoying:

```julia
using CuArrays
using CUDAdrv

println("Initial context:      ", repr(CUDAdrv.CuCurrentContext()))

CUDAdrv.Mem.alloc(1)

println("Context after init:   ", CUDAdrv.CuCurrentContext())

handle = CuArrays.CUBLAS.cublasXtCreate()

println("Context after create: ", CUDAdrv.CuCurrentContext())

CuArrays.CUBLAS.cublasXtDeviceSelect(handle, 1, Cint[0])

println("Context after select: ", CUDAdrv.CuCurrentContext())
```

```
Initial context:      nothing
Context after init:   CuContext(Ptr{Nothing} @0x0000000002daa2c0, true, true)
Context after create: CuContext(Ptr{Nothing} @0x0000000002daa2c0, true, true)
Context after select: CuContext(Ptr{Nothing} @0x0000000002323a00, false, true)
```

```c++
#include <stdio.h>

#include <cublasXt.h>
#include <cuda_runtime.h>
#include <cuda.h>

int main(int argc, char **argv) {
  CUcontext ctx = NULL;
  cuCtxGetCurrent(&ctx);
  printf("Initial context:      %p\n", ctx);

  void *dummy;
  cudaMalloc(&dummy, 1);

  cuCtxGetCurrent(&ctx);
  printf("Context after init:   %p\n", ctx);

  cublasXtHandle_t handle;
  if (cublasXtCreate(&handle) != CUBLAS_STATUS_SUCCESS) {
    fprintf(stderr, "!!!! CUBLASXT initialization error\n");
    return EXIT_FAILURE;
  }

  cuCtxGetCurrent(&ctx);
  printf("Context after create: %p\n", ctx);

  int devices[] = {0};
  if (cublasXtDeviceSelect(handle, 1, devices) != CUBLAS_STATUS_SUCCESS) {
    fprintf(stderr, "!!!! CUBLASXT device selection error\n");
    return EXIT_FAILURE;
  }

  cuCtxGetCurrent(&ctx);
  printf("Context after select: %p\n", ctx);

  if (cublasXtDestroy(handle) != CUBLAS_STATUS_SUCCESS) {
    fprintf(stderr, "!!!! shutdown error (A)\n");
    return EXIT_FAILURE;
  }
}
```

```
Initial context:      (nil)
Context after init:   0x555f3e254c20
Context after create: 0x555f3e254c20
Context after select: 0x555f3e254c20
```

maleadt:
OK, seems like we need to use the primary context mechanism. Looking into this.

kshyatt:
What the heck. Am I reading that right that it only mutates stuff in the Julia version, and only after device selection? Weeeeeeird.

I'm "glad" my suspicion about context funny business turned out to be right 😛

maleadt:
> What the heck. Am I reading that right that it only mutates stuff in the Julia version, and only after device selection? Weeeeeeird.

Yeah, it seems like cublasXt uses the active _primary_ context, and allocates a new one if it doesn't exist. We were using regular contexts, so initializing cublasXt replaced our context with a primary one.
https://github.com/JuliaGPU/CUDAnative.jl/pull/360 should fix this.

maleadt:
This passes tests locally, back to you @kshyatt :slightly_smiling_face: 
bors try

kshyatt:
wooo! DO NOT merge this, I have more tests to write

kshyatt:
bors try

kshyatt:
OK, I think this is mergeable if others think it looks good.

test/blas.jl
```diff
@@ -1380,6 +1379,298 @@ end
     @test C ≈ h_C
 end
 
-end
+@testset "xt" begin
+    @testset "trmm! with element type $elty" for elty in [Float32, Float64, ComplexF32] # test broken for ComplexF64
```
vchuravy: What part is broken here? Is it a tolerance issue?

kshyatt: Some of the off diagonal elements are wrong (and not by a little), I think.

src/blas/CUBLAS.jl
```diff
@@ -35,6 +37,24 @@ function handle()
...
 end
 
+function xt_handle()
+    if _xt_handle[] == C_NULL
+        @assert isassigned(active_context) # some other call should have initialized CUDA
+        _xt_handle[] = get!(_xt_handles, active_context[]) do
+            context = active_context[]
+            dev = CUDAdrv.device(context)
+            handle = cublasXtCreate()
+            cublasXtDeviceSelect(handle, 1, [dev.handle])
```
maleadt: why not default to using all available devices? isn't that the whole point of cublasXt?
it seems like you can't switch modes afterwards (or at least not between calls), so it'd need to happen here.

>  Currently the device configuration is static and cannot be changed between Math function calls. In that regard, this function should be called only once after cublasXtCreate. To be able to run multiple configurations, multiple cublasXt API contexts should be created.

test/blas.jl
kshyatt: I tried this locally and it works now??? Weird.

src/blas/CUBLAS.jl
```diff
@@ -35,6 +37,24 @@ function handle()
...
+            #cublasXtSetBlockDim(handle, 64)
```
maleadt: I guess this can go? This call can happen later.

kshyatt:
bors try

test/sparse_solver.jl
```diff
@@ -81,6 +81,7 @@ k = 1
 
     @testset "csreigvsi" begin
         A     = sparse(rand(elty,n,n))
+        A     = A + A'
```
maleadt: didn't tests pass before this change?

test/runtests.jl
```diff
@@ -22,6 +21,7 @@ testf(f, xs...; kwargs...) = GPUArrays.TestSuite.compare(f, CuArray, xs...; kwar
 
 allowscalar(false)
 
+using Test
```
maleadt: maybe get rid of this change (unless it has a reason?)

test/sparse_solver.jl
kshyatt: Not on my machine

src/blas/CUBLAS.jl
kshyatt: I set this for testing purposes and forgot about it! I'll try adding in more devices too :)

kshyatt:
bors try

kshyatt:
Yaaaaay! Thanks, @maleadt!

PR #308, opened 2019-03-28
more tests and missing methods for CURAND

kshyatt:


kshyatt:
bors try

test/rand.jl
```diff
@@ -47,6 +49,10 @@ for (f,T) in ((rand!,Int64),),
     f(A)
 end
 
+@test_throws ErrorException CURAND.normal_rng(CuArray{Cuint}(undef, 10))
```
maleadt: No need to check these using the internal API, you can trigger the exception using exported functions:

```
julia> CURAND.rand_poisson!(CuArray{Float64}(undef, 10))
ERROR: CuArrays does not support generating Poisson distributed numbers of type Float64
```

Issue #11, opened 2017-09-01
transpose is much slower than permutedims


ViralBShah:
`permutedims(x, (2,1))` works really fast. `transpose` does not. May just need to be hooked up.

MikeInnes:
`transpose` was implemented in #29.

PR #330, opened 2019-04-29
cuTENSOR wrappers

kshyatt:
You'll need to put `libcutensor.so` in a reasonable place. I ran all these tests separately but we should check they actually work integrated into CuArrays too.


kshyatt:
This also includes some high-level type wrapping too so you can just add tensors and stuff. We can integrate stuff like `mapslices` and broadcasting later imo.

maleadt:
Great! I'll have a closer look as soon as I have some time.

cc @Jutho @springer13

src/tensor/libcutensor_types.jl
```diff
@@ -0,0 +1,96 @@
...
+const CUTENSOR_ALGO_LOG            = Int32(-4)
+const CUTENSOR_ALGO_TTGT_TENSOR_OP = Int32(-3)
+const CUTENSOR_ALGO_TTGT           = Int32(-2)
+const CUTENSOR_ALGO_DEFAULT        = Int32(-1)
+
+const cutensorTensorDescriptor_t = Ptr{Cvoid}
+const cutensorContext = Cvoid
+const cutensorHandle_t = Ptr{cutensorContext}
+    
+const cudaDataType_t = UInt32
```
maleadt: These live in CUDAapi now: https://github.com/JuliaGPU/CUDAapi.jl/blob/0ec114a4ae2eecbecf05963532123f4ba56dbaf3/src/library_types.jl#L7-L22

src/tensor/wrappers.jl
```diff
@@ -0,0 +1,315 @@
+using CUDAdrv: CuDefaultStream, CuStream, CuStream_t
+
+function cudaDataType(T::DataType)
```
maleadt: This would make sense to move into CUDAapi instead, as it's probably used by CUBLAS too.

Jutho:
Looks like we've been doing a bit of double work. Anyway, great. I won't have much time until the semester is over (three more weeks), so it's good to see this being done. I guess it wouldn't then cost too much time finalize and release my CuTensorOperations.jl package which builds on top of CuTensor and TensorOperations.jl.

maleadt:
> Looks like we've been doing a bit of double work.

FYI, this is based on the latest versions of maleadt/CUTENSOR.jl and Jutho/CuTensor.jl.

Jutho:
Ah; ok. I didn't look at it in detail, so I did indeed not appreciate that :-).

Jutho:
However, cutensor has undergone  some changes since then, and I haven't had time to update the Julia wrappers. I think it's mostly the contraction routine, which now accepts a work buffer as argument. @springer13 can comment.

src/tensor/wrappers.jl
kshyatt: We aren't using it in CUBLAS yet because we don't have tests or anything for mixed precision gemmed

maleadt:
Pushed some fixed, but can't seem to get this to work (CUDA and CUBLAS errors).
The errors reproduce with the samples that are shipped with cutensor:

```
$ ./elementwise_binary
Total memory: 0.18 GiB
ERROR: CUTENSOR_STATUS_CUDA_ERROR
ERROR: CUTENSOR_STATUS_CUDA_ERROR
ERROR: CUTENSOR_STATUS_CUDA_ERROR
cuTensor: 13880.85 GB/s
memcpy: 166.82 GB/s
```

@kshyatt which versions did you use? I'm testing CUTENSOR 0.1.10 with CUDA 10.1.
@springer13 are there specific version requirements? (if so, it would be nice to have a `cutensorGetProperty` to query and enforce a version).

maleadt:
Updated to CUTENSOR 0.1.14 on CUDA 10.1, same error:

```
$ ./elementwise_binary                                                                                                                                                                                          
Total memory: 0.18 GiB
ERROR: CUTENSOR_STATUS_CUDA_ERROR
ERROR: CUTENSOR_STATUS_CUDA_ERROR
ERROR: CUTENSOR_STATUS_CUDA_ERROR
cuTensor: 13356.83 GB/s
memcpy: 161.90 GB/s
```

Jutho:
I will look at this as soon as I finished reading your thesis (almost done) :-).

springer13:
Hi all, sorry for being late to this conversation (need to update my mail address). It is great to see all this progress.
The latest cuTENSOR version requires SM60 (or later) as well as the CUDA 10.1 toolkit and the 418.xx driver. We'll provide more builds in the future.
@maleadt I assume that you'd like cutensorGetProperty() to behave similar to its cuBLASLt counter-part? You could look into the cutensor.h header it has three defines that encode the current version. Is this enough for now?
Please let me know if you can resolve the CUDA_ERROR.

maleadt:
`SM60` is the missing one, I'll test on a more recent GPU soon.

>  @maleadt I assume that you'd like cutensorGetProperty() to behave similar to its cuBLASLt counter-part?

Yes. I assumed that the `libraryPropertyType_t` mechanism would be standard across CUDA libraries (and it's implemented by many already).

> You could look into the cutensor.h header it has three defines that encode the current version. Is this enough for now?

No, since we don't use the header, we just call directly into the library so we need to be able to query for that kind of information.

springer13:
Okay thanks, I'll put this on my TODO list. We should include this as part of our next minor release.
Please let me know if you need anything else regarding the API.

Jutho:
I have everything running (after some upgrade trouble of the CUDA toolkit). Only the `elementwiseTrinary` test fails. Not sure what is going on there, just in the case of simple matrices and with A and C nonzero and B zero, it yields the correct result for D = A + B + C, but with B non-zero, a different result is obtained. With only B nonzero, and A and C zero, `D` contains entries of the order 10^{-25} (in Float32).

I don't see an immediate problem with the function definitions though, it seems the B array is passed along correctly. 

Of course, even when the above is fixed, there need to be many more tests (different eltypes etc). I'll make some time this this week.

kshyatt:
Yes, more tests! I can help with that too, I wanted to get some barebones
in to make sure I wrote the wrappers properly in the first place.

kshyatt:
Added some more test types and commented out (for now) the weird trinary fail. I'll try to add in combining integer and float types in addition etc.

Jutho:
That's great. I think that in my earlier explorations with the cutensor library, I never tested or even used/wrapped the tirnary function. Could it be that there is really an error in the implementation in the CUTENSOR library itself, @springer13 ?

maleadt:
And we have CI! Which fails though, error 18 (unknown CUDA error) with CuArrays.jl and error 20 (insufficient driver version) when running the CUDA C samples. Even though I do have 418.xx and CUDA 10.1... @springer13 ?

```
# ./contraction
Total memory: 0.45 GiB
Error: CUTENSOR_STATUS_INSUFFICIENT_DRIVER in line 156

# ./elementwise_binary 
Total memory: 0.18 GiB
CUTENSOR ERROR: some argument is NULL.
ERROR: CUTENSOR_STATUS_INVALID_VALUE
CUTENSOR ERROR: some argument is NULL.
ERROR: CUTENSOR_STATUS_INVALID_VALUE
CUTENSOR ERROR: some argument is NULL.
ERROR: CUTENSOR_STATUS_INVALID_VALUE
cuTensor: 76534.50 GB/s
memcpy: 810052.77 GB/s
terminate called after throwing an instance of 'cutensor::InternalError'
  what():  cublasDestroy failed.

Aborted (core dumped)
```

```
# nvcc --version
nvcc: NVIDIA (R) Cuda compiler driver
Copyright (c) 2005-2019 NVIDIA Corporation
Built on Wed_Apr_24_19:10:27_PDT_2019
Cuda compilation tools, release 10.1, V10.1.168

$ nvidia-smi
Thu May 23 11:52:43 2019       
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 418.56       Driver Version: 418.56       CUDA Version: 10.1     |
|-------------------------------+----------------------+----------------------+

```

CUTENSOR 0.1.14

maleadt:
This looks like an nvidia-docker issue? Or at least the combination of cutensor and nvidia-docker goes wrong:

Host system (CUDA 10.1, driver 418.56):
```
cutensor/samples$ CPATH=../include LIBRARY_PATH=../lib make -j10

cutensor/samples$ LD_LIBRARY_PATH=../../../lib ./contraction
Total memory: 0.45 GiB
cuTensor: 13663.96 GFLOPs/s 14.09 GB/s
```

Using docker:
```
$ docker run --rm -it -v $(pwd):/cutensor nvidia/cuda:10.1-devel

cutensor/samples# CPATH=../include LIBRARY_PATH=../lib make -j10

cutensor/samples# LD_LIBRARY_PATH=../../../lib ./contraction
Total memory: 0.45 GiB
Error: CUTENSOR_STATUS_INSUFFICIENT_DRIVER in line 156

```

Jutho:
@kshyatt , did you also observe that the `elementwiseBinary` tests fail specifically for the `ComplexF16` case, and work for all other `eltype`s?

springer13:
@Jutho Could you please provide me an exemplary input that I could use to reproduce the trinary issue?
@maleadt I'll double-check how other CUDA libraries report errors in the absence of a CUDA runtime and report errors accordingly.

Jutho:
I've updated the `elementwiseBinary` test to test more operations (both unary and binary) and a varying number of indices `N=2:5`. Let me know what you think; if you all agree I will generalise the other tests in a similar fashion.

@springer13 , so far, I've seen errors in `elementwiseBinary` with mixed element types and with `CUTENSOR_OP_SQRT` when the elements are complex. Is this to be expected?

 I've also changed `CharUnion`, used in the argument type for specifying the modes, to accept arbitrary `Integers`. I don't know if this is fair or common; the original type in cutensor is `Int32`. I would certainly appreciate to accept at least that, and not only characters.

maleadt:
Looks like CI is working! The failures are unrelated, and have started occurring after upgrading CUDA for this PR. I'll look into that.

springer13:
@Jutho What type of errors? I'd say that a NOT_SUPPORTED status would be expected since we have not instantiate all different data type combinations (for instance, sqrt of complex is not supported).

Jutho:
@maleadt , are you sure you're changes to `zeros` calls `CuArrays.zeros` and not `Base.zeros`. I suddenly get
```
  MethodError: no method matching unsafe_convert(::Type{CUDAdrv.CuPtr{Nothing}}, ::Array{Complex{Float32},1})
```
everywhere

Jutho:
I've changed from `(cu)zeros` to `similar` for the workspace of `contraction!`; I think that's good enough, no need for zero-initialized memory. I could also just use the `CuArray` constructor probably, it should anyway be device memory.

I've also generalised the permutation and contraction tests. I will look at ternary tests and provide a full report of errors encountered when enabling all possible test cases (many of which are currently commented out because they fail).

maleadt:
@Jutho yeah I forgot about the CUTENSOR submodule (didn't have the means to test locally). But `similar` is even better indeed, thanks for the fix.

Jutho:
I also removed the `CuTensor <: AbstractArray` subtype rule, as `CuTensor` has no `getindex` defined, and I don't think this is needed anywhere. I could see arguments both pro and contra making `CuTensor` a subtype, so I am open to revisit, but the way it currently behaves under e.g. `collect` seems incompatible with how an `AbstractArray` is supposed to behave.

Jutho:
@springer13  and @maleadt , I was experimenting with the experimental feature to pass arrays living in the host memory to cutensor functions, in particular `cutensorElementwiseBinary` as a first attempt. This requires, in the current implementation, a change in the wrapper functions, which I've just pushed as latest commit.

However, I observe strange behaviour. Running a test case once with host memory arrays works fine. However, after this one test, trying to copy a new (completely newly created array) from host to device gives the following error:
```
julia> A = rand(eltyA, dimsA...)
julia> dA = CuArray(A)
ERROR: CUDA error: an illegal memory access was encountered (code #700, ERROR_ILLEGAL_ADDRESS)
Stacktrace:
 [1] macro expansion at /home/jhaegemn/.julia/dev/CUDAdrv/src/base.jl:147 [inlined]
 [2] #copy!#11(::Nothing, ::Bool, ::Function, ::CUDAdrv.Mem.DeviceBuffer, ::Ptr{Float64}, ::Int64) at /home/jhaegemn/.julia/dev/CUDAdrv/src/memory.jl:344
 [3] copy! at /home/jhaegemn/.julia/dev/CUDAdrv/src/memory.jl:335 [inlined]
 [4] copyto!(::CuArray{Float64,2}, ::Int64, ::Array{Float64,2}, ::Int64, ::Int64) at /home/jhaegemn/.julia/dev/CuArrays/src/array.jl:186
 [5] copyto! at /home/jhaegemn/.julia/packages/GPUArrays/pJw1Y/src/abstractarray.jl:101 [inlined]
 [6] CuArray{Float64,2}(::Array{Float64,2}) at /home/jhaegemn/.julia/dev/CuArrays/src/array.jl:106
 [7] CuArray(::Array{Float64,2}) at /home/jhaegemn/.julia/dev/CuArrays/src/array.jl:116
 [8] top-level scope at none:0
```
Can it be that this experimental feature corrupts something?

Jutho:
@springer13 , another error which I have observed
Error 14 (internal operation failed) for `elementwiseBinary` with `Complex{Float16}` datatype.

Jutho:
Also, in contraction, the only mixed case that seems to work is
```julia
(eltyA, eltyB, eltyC) = (Float16, Float16, Float32)
```
even though cutensor prints an error:
```
CUTENSOR ERROR: Combination of provided data types is not yet supported; please let us know about this use case.
```
but still computes the result correctly.

Jutho:
@springer13 , with `elementwiseTernary`, it seems that `beta` is always assumed zero, i.e. the second array `B` does not contribute. The result obtained corresponds to `alpha * (permuted A) + gamma*C` independent of `B` or `beta`. This works for homogeneous datatype `Float32`, `Float64`, `ComplexF32` and `ComplexF64`. However, it fails for homogeneous datatype `Float16` (without printing an error).

Jutho:
Wait, my apologies for the noise, this seems to be an error in the julia wrapper of that function.

Jutho:
Fixing this seems to make `elementwiseTrinary` work. I've just pushed the implementation changes, I'll push the changes to the tests later tonight; taking a break for now.

Jutho:
@springer13 , I am also a bit confused about which contractions are supported with `Float16` datatype. I don't know if the following is the correct variable to put this out against, but I currently run tests with number of uncontracted indices of `A` and `B` (`NoA` and `NoB`) and number of contracted indices (`Nc`) each running from 1 to 3, and this yields the following:

- [x] NoA = 1, NoB = 1, Nc = 1
- [x] NoA = 1, NoB = 1, Nc = 2
- [ ] NoA = 1, NoB = 1, Nc = 3
- [x] NoA = 1, NoB = 2, Nc = 1
- [ ] NoA = 1, NoB = 2, Nc = 2
- [ ] NoA = 1, NoB = 2, Nc = 3
- [x] NoA = 1, NoB = 3, Nc = 1
- [ ] NoA = 1, NoB = 3, Nc = 2
- [ ] NoA = 1, NoB = 3, Nc = 3
- [x] NoA = 2, NoB = 1, Nc = 1
- [ ] NoA = 2, NoB = 1, Nc = 2
- [ ] NoA = 2, NoB = 1, Nc = 3
- [ ] NoA = 2, NoB = 2, Nc = 1
- [ ] NoA = 2, NoB = 2, Nc = 2
- [x] NoA = 2, NoB = 2, Nc = 3
- [ ] NoA = 2, NoB = 3, Nc = 1
- [ ] NoA = 2, NoB = 3, Nc = 2
- [x] NoA = 2, NoB = 3, Nc = 3
- [x] NoA = 3, NoB = 1, Nc = 1
- [x] NoA = 3, NoB = 1, Nc = 2
- [x] NoA = 3, NoB = 1, Nc = 3
- [x] NoA = 3, NoB = 2, Nc = 1
- [ ] NoA = 3, NoB = 2, Nc = 2
- [x] NoA = 3, NoB = 2, Nc = 3
- [x] NoA = 3, NoB = 3, Nc = 1
- [ ] NoA = 3, NoB = 3, Nc = 2
- [ ] NoA = 3, NoB = 3, Nc = 3

for the datatypes of `A`, `B` and `C` being `Float16`, `Float16`, `Float16` (or `Float16`, `Float16`, `Float32`, even though this prints a not-supported error as reported above, it does work in the selected case). For the following data types, all tests work.
```
                (Float32, Float32, Float32),
                (Float64, Float64, Float64),
                (ComplexF32, ComplexF32, ComplexF32),
                (ComplexF64, ComplexF64, ComplexF64)
```
Other cases don't work.

Jutho:
Ok, I am finished with this so far. I think all the functionality that currently works is more or less tested (not all unary operators like relu are tested), of course not in all possible combinations. But I think a wide range of cases is covered.

Remains to be done:
* testing error paths / types / messages
* fixing and testing the ability to pass host memory arrays (see above)
* documentation (here and for cutensor itself)

I am also not sure what the plans are with the `CuTensor` type, which I believe was implemetned by @kshyatt ? Should it be an `AbstractArray` subtype (currently no more). Should it support more methods? Should it be documented? Should there be something similar for host memory arrays?

The next few days/weeks, I will now spend my spare time on supporting cutensor from within TensorOperations.jl.

kshyatt:
Yeah that type was implemented by me as a convenience wrapper. I can add more methods to it so it can do more of what an abstract array should.

Jutho:
> > Looks like we've been doing a bit of double work.

> FYI, this is based on the latest versions of maleadt/CUTENSOR.jl and Jutho/CuTensor.jl.

There seem to be some significant changes in e.g. the construction of the `cutensorDescriptor`. If I see it correctly, they are currently never destroyed and thus the memory they occupy remains occupied? I created a new julia type for this that registered a finalizer so as to place a correct call to `cutensorDestroyTensorDescriptor`. Also, I need the ability to pass custom strides, so if everyone is ok, I will modify the implementation in closer spirit to what I had originally.

Jutho:
@maleadt , I am confused about the `PtrOrCuPtr` type. In the current implementation, I have separate wrapper methods (the functions that use `ccall`) for when the arrays are `Array` or `CuArray`, and I use `Ptr{Cvoid}` and `CuPtr{Cvoid}` as argument type respectively in `ccall`. I though to just simplify this to a single method with `PtrOrCuPtr{Cvoid}` as type, but this yields 
```
  ArgumentError: cannot take the CPU address of a GPU buffer
  Stacktrace:
   [1] convert(::Type{Ptr{Nothing}}, ::CUDAdrv.Mem.DeviceBuffer) at /home/jhaegemn/.julia/dev/CUDAdrv/src/memory.jl:83
   [2] unsafe_convert(::Type{Ptr{Nothing}}, ::CUDAdrv.Mem.DeviceBuffer) at /home/jhaegemn/.julia/dev/CUDAdrv/src/memory.jl:32
   [3] unsafe_convert(::Type{CUDAdrv.PtrOrCuPtr{Nothing}}, ::CUDAdrv.Mem.DeviceBuffer) at /home/jhaegemn/.julia/dev/CUDAdrv/src/pointer.jl:115
   [4] macro expansion at /home/jhaegemn/.julia/dev/CuArrays/src/tensor/error.jl:50 [inlined]
   [5] cutensorElementwiseBinary(::Ptr{Nothing}, ::Array{Complex{Float64},1}, ::CuArray{Complex{Float64},5}, ::CuArrays.CUTENSOR.CuTensorDescriptor, ::Array{Int32,1}, ::Array{Complex{Float64},1}, ::CuArray{Complex{Float64},5}, ::CuArrays.CUTENSOR.CuTensorDescriptor, ::Array{Int32,1}, ::CuArray{Complex{Float64},5}, ::CuArrays.CUTENSOR.CuTensorDescriptor, ::Array{Int32,1}, ::UInt32, ::CUDAapi.cudaDataType, ::CUDAdrv.CuStream) at /home/jhaegemn/.julia/dev/CuArrays/src/tensor/libcutensor.jl:58
```

Is this expected?

Jutho:
More changes to the implementation:
* reintroduced the `CuTensorDescriptor` type which registers a finalizer to destroy itself, in wrappers.jl
* revert restriction to device memory arrays, and thus to `CuPtr` type in `ccall` methods, until I know whether I can use `PtrOrCuPtr` and why this currently fails. The experimental feature of using host memory arrays was anyway broken (see above).
* move the `CuTensor` type to highlevel.jl

Now libcutensor.jl corresponds roughly to libcutensor.h, and libcutensor_types.jl to cutensor/types.h.  The file wrappers.jl corresponds to a more Julian approach to the methods in cutensor, and highlevel.jl introduces the `CuTensor` type and its associated methods.

springer13:
> Also, in contraction, the only mixed case that seems to work is
> 
> ```julia
> (eltyA, eltyB, eltyC) = (Float16, Float16, Float32)
> ```
> 
> even though cutensor prints an error:
> 
> ```
> CUTENSOR ERROR: Combination of provided data types is not yet supported; please let us know about this use case.
> ```
> 
> but still computes the result correctly.

Thanks for bringing this up; this might be a bug on our side. Could you please rerun with exporting the environment variable CUTENSOR_DEBUG=1 and post the output?

springer13:
> @springer13 and @maleadt , I was experimenting with the experimental feature to pass arrays living in the host memory to cutensor functions, in particular `cutensorElementwiseBinary` as a first attempt. This requires, in the current implementation, a change in the wrapper functions, which I've just pushed as latest commit.
> 
> However, I observe strange behaviour. Running a test case once with host memory arrays works fine. However, after this one test, trying to copy a new (completely newly created array) from host to device gives the following error:
> 
> ```
> julia> A = rand(eltyA, dimsA...)
> julia> dA = CuArray(A)
> ERROR: CUDA error: an illegal memory access was encountered (code #700, ERROR_ILLEGAL_ADDRESS)
> Stacktrace:
>  [1] macro expansion at /home/jhaegemn/.julia/dev/CUDAdrv/src/base.jl:147 [inlined]
>  [2] #copy!#11(::Nothing, ::Bool, ::Function, ::CUDAdrv.Mem.DeviceBuffer, ::Ptr{Float64}, ::Int64) at /home/jhaegemn/.julia/dev/CUDAdrv/src/memory.jl:344
>  [3] copy! at /home/jhaegemn/.julia/dev/CUDAdrv/src/memory.jl:335 [inlined]
>  [4] copyto!(::CuArray{Float64,2}, ::Int64, ::Array{Float64,2}, ::Int64, ::Int64) at /home/jhaegemn/.julia/dev/CuArrays/src/array.jl:186
>  [5] copyto! at /home/jhaegemn/.julia/packages/GPUArrays/pJw1Y/src/abstractarray.jl:101 [inlined]
>  [6] CuArray{Float64,2}(::Array{Float64,2}) at /home/jhaegemn/.julia/dev/CuArrays/src/array.jl:106
>  [7] CuArray(::Array{Float64,2}) at /home/jhaegemn/.julia/dev/CuArrays/src/array.jl:116
>  [8] top-level scope at none:0
> ```
> 
> Can it be that this experimental feature corrupts something?

Host pointers for any element-wise function is not supported; hence, the following error (i.e., ERROR_ILLEGAL_ADDRESS) is expected.

springer13:
> @springer13 , another error which I have observed
> Error 14 (internal operation failed) for `elementwiseBinary` with `Complex{Float16}` datatype.

ERROR_ILLEGAL_ADDRESS

> @springer13 , another error which I have observed
> Error 14 (internal operation failed) for `elementwiseBinary` with `Complex{Float16}` datatype.

Such an error should never occur. Could you please rerun with exporting the environment variable CUTENSOR_DEBUG=1 and post the output?

springer13:
> @springer13 , I am also a bit confused about which contractions are supported with `Float16` datatype. I don't know if the following is the correct variable to put this out against, but I currently run tests with number of uncontracted indices of `A` and `B` (`NoA` and `NoB`) and number of contracted indices (`Nc`) each running from 1 to 3, and this yields the following:
> 
> * [x]  NoA = 1, NoB = 1, Nc = 1
> * [x]  NoA = 1, NoB = 1, Nc = 2
> * [ ]  NoA = 1, NoB = 1, Nc = 3
> * [x]  NoA = 1, NoB = 2, Nc = 1
> * [ ]  NoA = 1, NoB = 2, Nc = 2
> * [ ]  NoA = 1, NoB = 2, Nc = 3
> * [x]  NoA = 1, NoB = 3, Nc = 1
> * [ ]  NoA = 1, NoB = 3, Nc = 2
> * [ ]  NoA = 1, NoB = 3, Nc = 3
> * [x]  NoA = 2, NoB = 1, Nc = 1
> * [ ]  NoA = 2, NoB = 1, Nc = 2
> * [ ]  NoA = 2, NoB = 1, Nc = 3
> * [ ]  NoA = 2, NoB = 2, Nc = 1
> * [ ]  NoA = 2, NoB = 2, Nc = 2
> * [x]  NoA = 2, NoB = 2, Nc = 3
> * [ ]  NoA = 2, NoB = 3, Nc = 1
> * [ ]  NoA = 2, NoB = 3, Nc = 2
> * [x]  NoA = 2, NoB = 3, Nc = 3
> * [x]  NoA = 3, NoB = 1, Nc = 1
> * [x]  NoA = 3, NoB = 1, Nc = 2
> * [x]  NoA = 3, NoB = 1, Nc = 3
> * [x]  NoA = 3, NoB = 2, Nc = 1
> * [ ]  NoA = 3, NoB = 2, Nc = 2
> * [x]  NoA = 3, NoB = 2, Nc = 3
> * [x]  NoA = 3, NoB = 3, Nc = 1
> * [ ]  NoA = 3, NoB = 3, Nc = 2
> * [ ]  NoA = 3, NoB = 3, Nc = 3
> 
> for the datatypes of `A`, `B` and `C` being `Float16`, `Float16`, `Float16` (or `Float16`, `Float16`, `Float32`, even though this prints a not-supported error as reported above, it does work in the selected case). For the following data types, all tests work.
> 
> ```
>                 (Float32, Float32, Float32),
>                 (Float64, Float64, Float64),
>                 (ComplexF32, ComplexF32, ComplexF32),
>                 (ComplexF64, ComplexF64, ComplexF64)
> ```
> 
> Other cases don't work.

what is the compute type in all these cases?

Jutho:
> what is the compute type in all these cases?

We use a simplified interface, where the computeType is always equal to the datatype of the output array, i.e. the last type in this list above.

Jutho:
> Such an error should never occur. Could you please rerun with exporting the environment variable `CUTENSOR_DEBUG=1` and post the output?

I did not manage to get further output printed. However, I now notice in the docs that there is no `CUDA_C_16F` support anyway, or at least this is not mentioned anywhere, so maybe that's an explanation.

Jutho:
> > Also, in contraction, the only mixed case that seems to work is
> > 
> > (eltyA, eltyB, eltyC) = (Float16, Float16, Float32)
> > even though cutensor prints an error:
> > 
> > CUTENSOR ERROR: Combination of provided data types is not yet supported; please let us know about this use case.
> > but still computes the result correctly.
> > 

> Thanks for bringing this up; this might be a bug on our side. Could you please rerun with exporting the environment variable CUTENSOR_DEBUG=1 and post the output?

```
A: -----------Tensor-------------
Number of modes: 2
Extents: 343 344 
Strides: 1 343 
Data type: 2
Op: 1
Vector-mode: 999
Vector-width: 1
------------------------------
modeA: 97,99,
B: -----------Tensor-------------
Number of modes: 2
Extents: 344 151 
Strides: 1 344 
Data type: 2
Op: 1
Vector-mode: 999
Vector-width: 1
------------------------------
modeB: 99,98,
C: -----------Tensor-------------
Number of modes: 2
Extents: 151 343 
Strides: 1 151 
Data type: 0
Op: 1
Vector-mode: 999
Vector-width: 1
------------------------------
modeC: 98,97,
Algo: -1
compute type: 0
A: -----------Tensor-------------
Number of modes: 2
Extents: 343 344 
Strides: 1 343 
Data type: 2
Op: 1
Vector-mode: 999
Vector-width: 1
------------------------------
modeA: 97,99,
B: -----------Tensor-------------
Number of modes: 2
Extents: 344 151 
Strides: 1 344 
Data type: 2
Op: 1
Vector-mode: 999
Vector-width: 1
------------------------------
modeB: 99,98,
C: -----------Tensor-------------
Number of modes: 2
Extents: 151 343 
Strides: 1 151 
Data type: 0
Op: 1
Vector-mode: 999
Vector-width: 1
------------------------------
modeC: 98,97,
Algo: -1
compute type: 0
CUTENSOR ERROR: Combination of provided data types is not yet supported; please let us know about this use case.
```

springer13:
Yes, the only supported mixed-precision implementations --for tensor contractions-- are those that are listed in cutensor.h (i.e., same input and output type but different compute type). Please let me know if you are interested in support for other combination.s

springer13:
> > Such an error should never occur. Could you please rerun with exporting the environment variable `CUTENSOR_DEBUG=1` and post the output?
> 
> I did not manage to get further output printed. However, I now notice in the docs that there is no `CUDA_C_16F` support anyway, or at least this is not mentioned anywhere, so maybe that's an explanation.

Thanks for bringing this up. The next version of cuTENSOR will correctly report this use case as NOT_SUPPORTED.

springer13:
> Also, in contraction, the only mixed case that seems to work is
> 
> ```julia
> (eltyA, eltyB, eltyC) = (Float16, Float16, Float32)
> ```
> 
> even though cutensor prints an error:
> 
> ```
> CUTENSOR ERROR: Combination of provided data types is not yet supported; please let us know about this use case.
> ```
> 
> but still computes the result correctly.

Could you please provide me with the output after setting CUTENSOR_DEBUG=1 ? I was not able to reproduce this issue so far.

Jutho:
My last post above has the output for this case.

Jutho:
I've added support for the latest version of cutensor (0.2.2) by implementing and testing the reduction interface, and also testing the mixed real times complex contraction. I've added Libdl to the project.toml.

Someone (@maleadt ?) should also look at the new support to query version numbers, and of course to make the tests use the latest version of cutensor.

Jutho:
? I guess this was unintentionally closed by merging CUDAdrv.jl#145 ?

maleadt:
Oh yeah that wasn't intended.

kshyatt:
OK, I rebased this on top of master and the tests passed locally...

kshyatt:
bors try

kshyatt:
bors try

kshyatt:
We're getting rekt by incompatible compute types it looks like...

Jutho:
Is this using the latest version of cutensor ? That should not happen; you also confirmed that the tests passed locally @kshyatt ?

kshyatt:
They did pass locally for me. I suspect something got overwritten in the rebase...

kshyatt:
Since this hadn't been run in almost two weeks, could something have happened to the docker image? I'm finding it hard to see where in the PR this could be going wrong...

maleadt:
Which version of cuTENSOR are you using? CI doesn't automatically download it, it's the binary I pushed onto it a couple of months back.

maleadt:
Upgraded the library, but it is linked against CUDA 10.0 while the CI is now running CUDA 10.1. 
@springer13 mind rebuilding the library and pushing a version?

maleadt:
bors try

Jutho:
Any progress here? I noticed yesterday that 10.1 compatible binaries are now available from https://developer.nvidia.com/cuTensor .

kshyatt:
Yes, @maleadt, @springer13 and I spent some time wrangling linker issues. We just need to update the Docker image with the new `.so` file, but Tim is on vacation.

maleadt:
OK, I'm back. Updated the container with the latest cuTENSOR 0.2.2 for CUDA 10.1.
bors try

Probably won't work, because the release still seems broken, linking `libcutensor` against `libcublas.so.10.0` while CUDA 10.1 provides `libcublas.so`, `libcublas.so.10` and `libcublas.so.10.2.0.168`. This looks like the same issue as when I looked at it during JuliaCon, @kshyatt?

kshyatt:
When I tried the new version on our cluster it seemed to be linking to the correct thing ... I can check again later today.

maleadt:
Seems like I downloaded the wrong version.
bors try

maleadt:
OK, rebased manually. Still needs the improvements from https://github.com/JuliaGPU/CuArrays.jl/pull/381 for CUTENSOR, but not straightforward due to the init-time libdl shenanigans.

maleadt:
Yay! To prevent this from regressing, the tests should probably require all submodules (CUDNN, CUTENSOR) to be tested if `haskey(ENV, "CI")`.

maleadt:
Is this good to go?

kshyatt:
Rebased to fix the conflict.

kshyatt:
bors try

maleadt:
I have removed the Libdl fiddling, it didn't seem necessary on my system. Maybe something has changed with the cutensor build process? Either way, @Jutho, you originally added that, care to see if it still works on your system? Besides that, this should be good to go I think.

maleadt:
CI looks good, let's merge this!

PR #344, opened 2019-05-14
Add more methods for seed!

findmyway:
Add more methods for seed! to keep it compatible with [Random.seed!](https://github.com/JuliaLang/julia/blob/master/stdlib/Random/src/RNGs.jl#L288-L290)

### DIFF
--- a/src/rand/CURAND.jl
+++ b/src/rand/CURAND.jl
@@ -43,6 +43,7 @@ version() = VersionNumber(curandGetProperty(CUDAapi.MAJOR_VERSION),
 
 end
 
+const seed! = CURAND.seed!
 const rand = CURAND.rand
 const randn = CURAND.randn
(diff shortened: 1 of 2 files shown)


maleadt:
Thanks! Need to look into this a little closer to see if it checks all the boxes, but this should probably also be exported so that you can do `CuArrays.rand!` (https://github.com/JuliaGPU/CuArrays.jl/blob/2a401e245ce8ce2659b973323afb9a5cd485fcc9/src/rand/CURAND.jl#L45-L48). Also, we need to think what to do with the GPUArrays RNG, which is used for types that aren't supported by CURAND; `seed!` should probably seed both.

maleadt:
Sorry for letting this fall behind. Rebased; thanks for the contribution!
bors r+

PR #352, opened 2019-06-04
More tests for BLAS and improve error msg

kshyatt:


kshyatt:
You need that `BLAS.dotc` to get the pass-through to work.

kshyatt:
bors try

kshyatt:
bors' fail is in FFT

PR #358, opened 2019-06-14
Add dims support to similar for broadcasted arrays

lcw:
Note, the tests for this functionality have been added to the GPUArrays
test suite.


simonbyrne:
Can we tag a new release?
cc @vchuravy

PR #366, opened 2019-07-11
Work around / ignore cuFFT issues.

maleadt:
Try to fix https://github.com/JuliaGPU/CuArrays.jl/issues/345

Something iffy is going on here, it looks like CUFFT changes inputs that it really shouldn't:

```julia
function fail1()
    T = Float32

    Random.seed!(1)
    X = rand(T, 2, 2)

    Y = rfft(X)
    d_X = CuArray(X)
    d_Y = CuArray(Y)

    pinv = plan_brfft(d_Y,size(X,1))
    bak = copy(d_Y)
    d_Z = pinv * d_Y
    @assert Array(d_Y) == Array(bak)
end

# cufftExecC2R in fft/libcufft.jl: @show idata round ccall
```
```
idata = Complex{Float32}[2.49235+0.0im 0.242207+0.0im; -0.26089+0.0im -0.281692+0.0im]
idata = Complex{Float32}[2.73456+0.0im 2.25014+0.0im; -0.542582+0.0im 0.020802+0.0im]
```

@kshyatt, I think you wrote most of these wrappers, any ideas? It should be out-of-place as per the docs, as `idata != odata`. Looks like other people have similar issues though: https://devtalk.nvidia.com/default/topic/475992/cufftexecc2r-altering-input-data-after-call-input-data-changes-after-cufftexecc2r-is-called-for-non-/ https://devtalk.nvidia.com/default/topic/460195/cufft-out-of-place-transform-destroys-input-/

To make matters worse, replicating this issue is tricky. Doesn't reproduce on my system (CUDA 10.1, GTX 970), but does on the CI system (CUDA 10.1, GTX 2080 Ti), so it's hard to figure out when we need to copy. Doing so unconditionally seems unacceptable to me.

Remaining CUFFT issues:

```julia
function fail2()
    T = ComplexF32

    dims = (8, 32, 64, 8)
    for region in [(1,4)]
        X = rand(T, dims)

        fftw_X = fft(X,region)
        d_X = CuArray(X)
        p = plan_fft(d_X,region)
        d_Y = p * d_X
        Y = collect(d_Y)
        @test isapprox(Y, fftw_X, rtol = MYRTOL, atol = MYATOL)
    end
end
```

### DIFF
--- a/res/wrap/wrap.jl
+++ b/res/wrap/wrap.jl
@@ -105,13 +105,22 @@ mutable struct State
     edits::Vector{Edit}
 end
 
-# insert `@check` before each `ccall` when it returns a xxxStatus_t
+# insert `@check` before each `ccall` when it returns a checked type
+const checked_types = [
+    "cublasStatus_t",
+    "cudnnStatus_t",
+    "cufftResult",
+    "curandStatus_t",
+    "cusolverStatus_t",
+    "cusparseStatus_t",
+    "cutensorStatus_t",
+]
 function insert_check(x, state)
     if x isa CSTParser.EXPR && x.typ == CSTParser.Call && x.args[1].val == "ccall"
         # get the ccall return type
         rv = x.args[5]
 
-        if endswith(rv.val, "Status_t")
+        if rv.val in checked_types
             push!(state.edits, Edit(state.offset, "@check "))
         end
     end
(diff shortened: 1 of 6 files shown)


maleadt:
The remaining issue also only reproduces if I run on the CI server within Docker, even though that uses exactly the same version of CUDA and CUFFT. But using that same Docker image on my workstation doesn't trigger the issue...

maleadt:
Errors are CUSOLVER/CUTENSOR.

PR #368, opened 2019-07-12
Allow combined use of SubArray and PermutedDimsArray

maleadt:


maleadt:
Adapt has been tagged.
bors try

maleadt:
the `permutedims` test (which is pretty much the issue reported by @merckxiaan) works, so let's merge this.

PR #369, opened 2019-07-12
Fix ND GPUArrays.LocalMemory allocation.

maleadt:
A typical case of "how did this ever work" (`_shmem` takes a length argument).


maleadt:
Fixes the `transpose!` issue.

Issue #371, opened 2019-07-14
Allow CuArrays to initialise if there is no CUDA/GPU on the system


tanhevg:
Currently an attempt to `using CuArrays` on a system that has no CUDA/GPU will result in an error. As a consequence, if one has a module that uses `CuArrays`, and wants to use some of its functions on a system without a GPU (e.g. debug a model training loop on a toy data set on a laptop on a plane), a separate `.toml` without `CuArrays` has to be created. This can be avoided by wrapping a few lines in the initialiser in a simple try/catch.

maleadt:
Trying a different approach: https://github.com/JuliaGPU/CuArrays.jl/pull/375 and https://github.com/JuliaGPU/CUDAapi.jl/pull/84

PR #372, opened 2019-07-14
Allow CuArrays to initialise if there is no CUDA/GPU on the system

tanhevg:
#371 
Wrap initialisation of active context in a try/catch

### DIFF
--- a/src/blas/CUBLAS.jl
+++ b/src/blas/CUBLAS.jl
@@ -4,7 +4,7 @@ import CUDAdrv: CUDAdrv, CuContext, CuStream_t, CuPtr, PtrOrCuPtr, CU_NULL, devi
 import CUDAapi
 
 using ..CuArrays
-using ..CuArrays: libcublas, active_context, unsafe_free!
+using ..CuArrays: libcublas, active_context, unsafe_free!, check_cudadrv_version
 using LinearAlgebra
 
 include("libcublas_types.jl")
--- a/src/blas/libcublas.jl
+++ b/src/blas/libcublas.jl
@@ -1837,7 +1837,7 @@ function cublasZtrttp(handle, uplo, n, A, lda, AP)
                handle, uplo, n, A, lda, AP)
 end
 
-if CUDAdrv.version() ≥ v"7.5"
+if check_cudadrv_version(v"7.5")
     # Wrap extensions of functions (ie. Nrm2Ex, GemmEx, etc) (CUDA 7.5+)
     function cublasNrm2Ex(handle, n, x, xType, incx, result, resultType, executionType)
         @check ccall((:cublasNrm2Ex, libcublas),
--- a/src/blas/libcublas_types.jl
+++ b/src/blas/libcublas_types.jl
@@ -92,7 +92,7 @@ const cublasXtPinningMemMode_t = UInt32
 const CUBLASXT_PINNING_DISABLED = 0
 const CUBLASXT_PINNING_ENABLED = 1
 
-if CUDAdrv.version() >= v"0.7.5"
+if check_cudadrv_version(v"0.7.5")
     # specify which GEMM algorithm to use in cublasGemmEx() (CUDA 7.5+)
     const cublasGemmAlgo_t = Int32
     const CUBLAS_GEMM_DFALT = -1
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -84,3 +84,5 @@ macro sync(ex)
         ret
     end
 end
+
+check_cudadrv_version(ver) = CUDAdrv.configured && CUDAdrv.version() >= ver
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -85,4 +85,4 @@ macro sync(ex)
     end
 end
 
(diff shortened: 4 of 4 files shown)


maleadt:
`configured` should be false if CUDA is not working, how come that isn't the case on your system?

tanhevg:
> `configured` should be false if CUDA is not working, how come that isn't the case on your system?

I guess this is because I based my PR on `v1.0.2`. With the latest changes on `master` the PR might be obsolete

tanhevg:
> `configured` should be false if CUDA is not working, how come that isn't the case on your system?

Yeah, I guess I have based the PR on the `v1.0.2` code. There were lots of changes in `master` that made it obsolete. Although now attempt to use `CuArrays#master` blows up if there is no CUDA/GPU, which brings us back to the "two sets of tomls" problem.

tanhevg:
Reverted the previous change, and made another one such that version checking does not blow up if there is no CUDA; please re-review.

maleadt:
Thanks. I'm currently restructuring how CUDA packages are loaded, so I'll be sure to incorporate these changes over there.

I'm wondering though, why do you want CuArrays to be loadable without CUDA/GPU? Would it be equivalent if only one of those packages, say CUDAapi, would be loadable without CUDA/GPU, providing functions like `has_cuda` and `has_gpu`? It would greatly simplify the package loading code. So concretely:

```julia
using CUDAapi # never fails
if has_cuda() && has_gpu()
  using CuArrays
end
```

cc @MikeInnes @ChrisRackauckas who have similar conditional module loading paths
(I'm trying to get rid of `Pkg.build` so that should further improve the conditional loading use case)

ChrisRackauckas:
CUDAapi really should have those functions. That as a standard setup would be really nice to have.

MikeInnes:
That seems like a good setup. If errors are only thrown when you try to use CuArrays, rather than at build time, that enables us to do things like using the CUDA automatically in Flux, which would be a big boon. I'm also not sure why you'd want to load CuArrays itself without the GPU though (and we have to at some point warn users that it isn't working, if they are expecting it to).

tanhevg:
> Would it be equivalent if only one of those packages, say CUDAapi, would be loadable without CUDA/GPU, providing functions like `has_cuda` and `has_gpu`? It would greatly simplify the package loading code.

Not sure I agree. All modules that use `CuArrays` would have to wrap all structures that reference `CuArray` (or any other type from `CuArrays` for that matter) in this `if has_cuda() && has_gpu()` clause. Otherwise, the module will not precompile. So with this approach we'll end up with anyone using `CuArrays` having to pepper their code with the if statements, which is not nice.

I think it is better for the module to be self-contained and loadable under a few well defined configurations, e.g. CUDA not installed, and CUDA installed properly. Especially, given that this is really not that hard to achieve.

MikeInnes:
Flux must be one of the more complex users of CuArrays, and even there it's pretty simple. Just wrap one `if` around an `include` statement that has the compat code (the same thing we do with Requires now). Even if it weren't simple, *someone* has to write that if statement, and I'm strongly in favour of doing that once in packages rather than making users do it.

Technical ease is not really the issue here; the harder design problem is how we avoid CUDA silently failing when a user was expecting it to work, and needs to be aware of an issue on their system. The later we push this, the more potential frustration for that group, which means we have ourselves a tradeoff.

ChrisRackauckas:
>Technical ease is not really the issue here; the harder design problem is how we avoid CUDA silently failing when a user was expecting it to work, and needs to be aware of an issue on their system. The later we push this, the more potential frustration for that group, which means we have ourselves a tradeoff.

Even when it doesn't fail, people are still getting warnings. CUDAnative really shouldn't be throwing a warning if there's no cudnn since only a small group actually cares about that, so the warning should be in a more targeted place. I think with the right if statements and if packages want to do fancy things, throw the right errors/warnings, everything will be clean. And so add `has_cudnn()` to the list and I think we have the right tools for packages to make it clean.

tanhevg:
>  I'm also not sure why you'd want to load CuArrays itself without the GPU though (and we have to at some point warn users that it isn't working, if they are expecting it to).

I just to write as little CUDA-specifict code as possible; prefer the modules that I use to do stuff for me.  Ditto if I am writing a module myself - prefer the `if` statement to be written once in `CuArrays`, than having to write it myself.

Currently, with this PR merged in, an attempt to `using CuArrays` on a machine without CUDA/GPU will produce a warning, and a call to the actual CUDA code will error. So I think we are fine on warnings/errors side.

maleadt:
> prefer the `if` statement to be written once in `CuArrays`, than having to write it myself

Except that you still need to add an `if` statement to your code, namely to check whether the now successfully-loaded `CuArrays` module is actually functional. It's also not a single check in CuArrays, since the package needs to check whether _its_ dependencies (CUDAdrv and CuArrays) are actually functional.

maleadt:
Superseded by https://github.com/JuliaGPU/CuArrays.jl/pull/375 and https://github.com/JuliaGPU/CUDAapi.jl/pull/84

PR #379, opened 2019-07-21
Support for limiting total memory usage.

maleadt:
https://stackoverflow.com/questions/57066480/limit-gpu-memory-in-julia-using-cuarrays

Just a quick feature for now, nothing permanent.

### DIFF
--- a/src/memory.jl
+++ b/src/memory.jl
@@ -247,9 +247,9 @@ end
 function try_cuda_alloc(bytes)
   # check the memory allocation limit
   if usage_limit[] !== nothing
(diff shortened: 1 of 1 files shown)


src/memory.jl
```diff
@@ -22,8 +22,17 @@ import Base.GC: gc
 
 const pool_lock = ReentrantLock()
 
+const usage_limit = Ref{Union{Nothing,Int}}()
+
 function __init_pool_()
   pool_timings!()
+
+  global usage_limit
```
KristofferC: Can be removed.

PR #381, opened 2019-07-25
Warn about missing libraries.

maleadt:
A warning seems annoying, but I don't think we can leave this one out (cc @MikeInnes, you've been vocal about such warnings).

### DIFF
--- a/src/CuArrays.jl
+++ b/src/CuArrays.jl
@@ -37,6 +37,7 @@ let
             Base.include_dependency(path)
         end
         @eval global const $lib = $path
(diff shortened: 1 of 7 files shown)


maleadt:
Or maybe the warning should only trigger when importing the submodule?

MikeInnes:
These generally come with CUDA anyway, right? In which case it seems fine.

The one annoying one might be CUDNN, where the warning might be annoying for people who don't need it. But I'm not sure where else we'd put that. (Maybe the first time you do a GPU convolution or something?) Not a blocker in any case.

maleadt:
> These generally come with CUDA anyway, right? In which case it seems fine.

No, this is only for CUDNN (and soon CUTENSOR); i.e. the ones that are optional. I'll have a look at trying to put it in the module import.

PR #389, opened 2019-08-10
added reverse for arbitrary dimension CuArrays

kraftpunk97:
```
julia> test_shapes = [
         [1, 2, 4, 3],
         [4, 2],
         [5],
         [2^5, 2^5, 2^5],
       ]
4-element Array{Array{Int64,1},1}:
 [1, 2, 4, 3]
 [4, 2]      
 [5]         
 [32, 32, 32]

julia> for testshape ∈ test_shapes
           a = Float32.(reshape(Vector(1:prod(testshape)), testshape...))
           a_ = cu(a)
           b = similar(a)
           b_ = similar(a_)

           println("testing for shape: $testshape")
           for i=1:length(testshape)
               b = reverse(a, dims=i)
               my_reverse(a_, b_, i)
               @test all(cu(b) .== b_)
               println("dimension $i is ok.")
           end
           
           println()
       end
testing for shape: [1, 2, 4, 3]
dimension 1 is ok.
dimension 2 is ok.
dimension 3 is ok.
dimension 4 is ok.

testing for shape: [4, 2]
dimension 1 is ok.
dimension 2 is ok.

testing for shape: [5]
dimension 1 is ok.

testing for shape: [32, 32, 32]
dimension 1 is ok.
dimension 2 is ok.
dimension 3 is ok.
```


kraftpunk97:
benchmarks...
```
shape_ = [2^10, 2^10];
N = 2^10
a = Float32.(reshape(Vector(1:prod(shape_)), shape_...))# |> cu
a_ = cu(a)
b = similar(a)
b_ = similar(a_)

julia> @btime b =reverse(a, dims=2)
  740.175 μs (6 allocations: 4.00 MiB)

julia> @btime my_reverse(a_, b_, 2)
  129.330 μs (51 allocations: 1.83 KiB)
```

src/array.jl
```diff
@@ -327,3 +327,60 @@ function Base.reverse(v::CuVector, start=1, stop=length(v))
...
+    nthreads = 256
+    nblocks = ceil(Int, length(input)/nthreads)
+    shmem = nthreads * sizeof(T)
+
+    shape = [size(input)...]
+
+    numelemsinprevdims = prod(shape[1:dims-1])
+    numelemsincurrdim = shape[dims]
+
+    function kernel(input::CuDeviceArray{T, N}, output::CuDeviceArray{T, N}) where {T, N}
```
vchuravy: Why is this using shared memory?

If I simplify this a bit it looks like you are doing:
```
shared[idx] = in[I]
J = #= Index transformation =#
out[J] = shared[idx]
```

kraftpunk97: CUDA beginner here. So, can I do this without using shared memory?

vchuravy: Yeah, you only need shared memory when you want to communicate between threads. Here you just fetch some data on one thread, calculate the index to where it goes, and then push it.

You can do this (and it maybe be faster) but that would mean you are doing a index transform on the shared data and each thread is writing back linearly.

So instead of doing:
|tid |1 | 2 | 3 | 4|
| --- | - | ---| -- |--- |
|in[I] | 11 | 12 | 13 | 14 |
|shared | 11 | 12 | 13 | 14 |
|shared | 11 | 12 | 13 | 14 |
|out[J] |11 | 12 | 13 | 14 |

You might do

|tid |1 | 2 | 3 | 4|
| --- | - | ---| -- |--- |
|in[I] | 11 | 12 | 13 | 14 |
|shared | 11 | 12 | 13 | 14 |
|reg | 14 | 13 | 12 | 11 |
|out[J] | 14 | 13 | 12 | 11 |

which might improve the coalescing of the memory operations for the write.

kraftpunk97: @vchuravy I decided to drop the idea of using shared memory all together. No point in adding complexity if it can be avoided. Did see significant speed up...

```
julia> @btime rev_d = my_reverse(A_d, dims=2)
  20.762 μs (60 allocations: 1.95 KiB)
```

kraftpunk97:
pinging @vchuravy

kraftpunk97:
pinging @maleadt. Had this pull request going stale. Please review and suggest further changes, if any. Would be awesome if this could get merged 🙂

maleadt:
Sorry for not responding. Pushed some small improvements.

@vchuravy we do need shared memory in the case `input` and `output` point to the same array (that is what we do with the other reverse kernel). That should probably be more clear by having two versions, one in-place with shared memory, and another out-of-place without.

I'm also wondering if we can't support the invocation with `dims` by using a view that selects the dimension in question.

maleadt:
@kraftpunk97 could you check "allow pushes from maintainers"?

kraftpunk97:
It's already selected.

maleadt:
Ugh, seems like I broke the PR. I'll open a new one...

PR #391, opened 2019-08-19
Define det(::CuQRPackedQ)

tkf:
This patch is the same as the PR I made for `LinearAlgebra`: https://github.com/JuliaLang/julia/pull/32887


andreasnoack:
Sorry, forgot that Bors is in charge here.

maleadt:
Yeah, this does seem to fail CI... https://gitlab.com/JuliaGPU/CuArrays.jl/-/jobs/275768797

maleadt:
The culprit is that `collect(d_F.Q)` does scalar indexing to materialize the `CuQRPackedQ`. Naively doing `collect(CuArray(d_F.Q))` or `Array(d_F.Q)` didn't work, so I've reverted the change for now.

PR #406, opened 2019-08-29
Create a CUDA context

dhairyagandhi96:
Thanks to `IRTools.jl`, we can do some nifty things with Julia IR. Like using a `dynamo` to walk through the deep IR and offload sensible ops to the GPU.

```julia
julia> c = Conv((3,3), 3 => 16, pad = (1,1), relu); # from Flux

julia> r = rand(Float32, 32, 32, 3, 100);

julia> cuda() do
           c(r)
       end # run on GPU

julia> a = rand(Float32, 5*10^4);

julia> b = rand(Float32, 5*10^4);

julia> cuda() do
           a + b
       end
50000-element Array{Float32,1}:
 0.9649581
 1.2122422
 0.423553
...
```

Notice the return type is a normal `Array`, meaning that without much fidgeting, it is trivial to offload computation to the GPU and continue where you left off.

There are a couple caveats, not all functions behave nicely yet and we need better test coverage, but opening it now to get some review and direction of the way forward


vchuravy:
Thanks! What is driving the choice to use IRTools over Cassette? I would prefer the maintenance burden to rest with Cassette (e.g. me)

dhairyagandhi96:
The choice was made for the little nicer control over the IR with IRTools. Also, it's conceptually simpler so maintaining it should be easier also.

It was also fairly straightforward to define in lesser code making it more readable. Mind you I'm no cassette pro, but definitely worth a discussion.

src/context.jl
```diff
@@ -0,0 +1,169 @@
...
+    ga = get_cached(array_bank, a)
+    gb = get_cached(array_bank, b)
+    cache(array_bank, $f(ga, gb))
+  end
+end
+
+function get_cached(array_bank, arr::Array{T,N})::CuArray{T,N} where {T,N}
+  haskey(array_bank, arr) ?
+    array_bank[arr] :
+    (array_bank[arr] = CuArray(arr))
```
MikeInnes: We could probably write `get_cached(cx, x)` as `cx[x]`.

src/context.jl
```diff
@@ -0,0 +1,169 @@
...
+end
+
+function (c::CUDACtx)(::typeof(broadcasted), f, args...)
```
MikeInnes: For `broadcast` I think we should leave the broadcast struct alone (so that CuArrays can't leak into the program), and instead do all conversion and computation in `materialize`.

src/context.jl
```diff
@@ -0,0 +1,169 @@
...
+  gargs = map(x -> get_cached(array_bank, x), args)
+  broadcasted(f, gargs...) |> x -> cache(array_bank, x)
+end
+
+function (c::CUDACtx)(::typeof(getproperty), o, s::Symbol)
```
MikeInnes: We shouldn't need this function, along with `getfield` etc; arrays should only converted when they are used as inputs to CUDA primitives. This would also return a `CuArray` to the users program.

src/context.jl
```diff
@@ -0,0 +1,169 @@
...
+
+wrap_cuize.((sum, similar, materialize))
+
+function (c::CUDACtx)(::typeof(reshape), arr, args...)
+  r = reshape(get_cached(arr), args...)
+  cache(array_bank, r)
+end
+
+@dynamo function (c::CUDACtx)(meta...)
+  meta == nothing && return
```
MikeInnes: This line is unnecessary (`meta` can never be nothing).

src/context.jl
```diff
@@ -0,0 +1,169 @@
...
+  ir = IR(meta...)
+  ir == nothing && return
+
+  pr = Pipe(ir)
```
MikeInnes: You could replace this code with `IRTools.recurse!(ir)` (there's some info in the docs if needed).

src/context.jl
```diff
@@ -0,0 +1,169 @@
...
+
+    pr[v] = Expr(:call, self, ex.args...)
+
+  end
+  return finish(pr)
+end
+
+# TODO: remove arbitrary things in one place
+get_cached(array_bank, t::Union{Type, UnitRange, Function, Broadcasted, Symbol, Module, Nothing, Missing, Ptr, CuPtr, T}) where {T <: Real} = t
+get_cached(array_bank, t::Union{Tuple, NamedTuple}) = map(get_cached, t)
```
MikeInnes: These should be removed (`get_cached` shouldn't be called on anything other than an array, when used as an input to a primitive)

src/context.jl
```diff
@@ -0,0 +1,169 @@
...
+
+get_cached(array_bank, x::CuArray) = x
```
MikeInnes: This would also hopefully never be called, since there should be no CuArrays in the original program.

src/context.jl
```diff
@@ -0,0 +1,169 @@
...
+
+function get_cached(x::T) where T
+  T <: Array && return get_cached(array_bank, x)
+  isstructtype(T) && return x
```
MikeInnes: More struct stuff we should get rid of. `get_cached` should error if the input isn't an array.

src/context.jl
```diff
@@ -0,0 +1,169 @@
...
+end
+
+"""
+  Disable `CUDACtx` for a function
+"""
+function noop_pass(f)
+  @eval (c::CUDACtx)(::typeof($f), args...) = $f(args...)
+end
+
+noop_pass.((get_cached, NNlib.check_spdf,
```
MikeInnes: Probably best if these are macros. It'd be nice to add an `@cuda` macro or similar for the purpose of overloading `CUDACtx`.

src/context.jl
```diff
@@ -0,0 +1,169 @@
...
+	))
+
+for f in names(NNlib)
```
MikeInnes: Better to do this explicitly per function.

src/context.jl
```diff
@@ -0,0 +1,169 @@
...
+  getfield(NNlib, f) isa Function || continue
+  @eval function (c::CUDACtx)(::typeof($f), args...)
+    gargs = map(get_cached, args)
+    cache(array_bank, $f(gargs...))
+  end
+end
+
+# Hold all the arrays related to the op
+# BitArray and friends would like an AbstractArray construct
+const array_bank = CUDACtx()
```
MikeInnes: We should use a fresh context whenever `cuda(...)` gets called. Right now a bunch of functions refer to the global context.

test/context.jl
```diff
@@ -0,0 +1,46 @@
+using CuArrays, Test
+using CuArrays.NNlib
+
+# Check simple ops work and broadcast
+@testset "simple ops" begin
+  W = rand(5, 5)
+  b = rand(5)
+  @test cuda(() -> W*b) ≈ W*b
```
MikeInnes: Good to check types here as well, e.g. that the output is still an `Array`.

MikeInnes:
@vchuravy there probably isn't much in it, so if the lead maintainers of this package strongly prefer Cassette then I imagine it'd be OK to port it over.

Though as Dhairya points out there's a couple of potential advantages to fine grained control of the IR pass; the main one is that it's easier to cut out classes of functions we're not interested in, e.g. intrinsics or certain modules in Base, avoiding some redundant recompilation.

src/CuArrays.jl
```diff
@@ -4,7 +4,7 @@ using CUDAapi, CUDAdrv, CUDAnative
 
 using GPUArrays
 
-export CuArray, CuVector, CuMatrix, CuVecOrMat, cu
+export CuArray, CuVector, CuMatrix, CuVecOrMat, cu, cuda
```
maleadt: seems like a pretty generic function to export (both `cu` and `cuda` are bound to confuse users). why not something that implies its action, e.g., `on_cuda`? or `@cuda` re `@async`?

src/context.jl
```diff
@@ -0,0 +1,169 @@
+using IRTools: isexpr, IR, @dynamo, postwalk
```
maleadt: maybe name this file (and wherever the term is used) `contextual.jl` re CUDAnative.jl's Cassette PR. CUDA already has context objects, so that might be confusing.

maleadt:
I guess we'll need some way to assert GPU execution to actually test this?

dhairyagandhi96:
Yeah, for the tests I was thinking just having a context which we can look into to assert that the array is actually in there and corresponds to memory associated with the GPU

src/context.jl
dhairyagandhi96: Yes, will get rid of it

src/context.jl
dhairyagandhi96: Done

src/context.jl
dhairyagandhi96: Thoughts on using `get!`? I suppose the extra network transfer would be a problem in this case.

src/context.jl
dhairyagandhi96: Done

src/context.jl
dhairyagandhi96: Done

dhairyagandhi96: Thanks for that!

src/context.jl
MikeInnes: You could look at `Base.@get!`, which avoids evaluating the new value if it's not needed.

src/CuArrays.jl
MikeInnes: I agree about not exporting this for now. In the longer term, if this is successful it should replace `cu` entirely (alongside all the other APIs, for most users), so a generic name seems appropriate.

I think `cuda() do ...` reads right, and provides an obvious space for options (`cuda(device=2) do ...`), but `@cuda` could work well too (especially in that it's a bit nicer for one liners).

vchuravy:
Grrml Gmail ate my reply:

Since CUDAnative will use Cassette and GPUifyLoops already does I would strongly prefer only having one tool in the GPU ecosystem to do this. I would be interested in making IRTools transforms/utility functions work with Cassette, which should work relatively straightforwardly.

src/contextual.jl
```diff
@@ -0,0 +1,152 @@
...
+    end
+    push!(ex, q)
+  end
+
+  quote
+    $(ex...)
+  end
+end
+
+@contextual :+ :- :* :/ sum similar materialize
```
MikeInnes: I think we should set these things up to explicitly call whatever lower-level bindings we have; it should show what it would look like if we got rid of `CuArray` altogether.

src/contextual.jl
```diff
@@ -0,0 +1,152 @@
...
+"""
+macro noop_pass(fs...)
+  ex = [:( (cx::CUDACtx)(::typeof($f), args...) = $f(args...) ) for f in fs]
+
+  quote
+    $(ex...)
+  end
+end
+
+@noop_pass get_cached NNlib.check_spdf
```
MikeInnes: Why do we need a noop for `get_cached`? That shouldn't ever be called in code that we're transforming, right?

src/contextual.jl
```diff
@@ -0,0 +1,152 @@
+using IRTools: isexpr, IR, @dynamo, postwalk
+using IRTools: meta, Pipe, finish, Variable, self
+using MacroTools: @forward
+
+import Base.Broadcast.broadcasted
```
MikeInnes: These imports are redundant now

PR #416, opened 2019-09-16
ND-reverse

maleadt:
Continuation of https://github.com/JuliaGPU/CuArrays.jl/pull/389

### DIFF
--- a/src/array.jl
+++ b/src/array.jl
@@ -378,8 +378,9 @@ function Base.reverse(input::CuArray{T, N}; dims::Integer) where {T, N}
         index_in = offset_in + threadIdx().x
 
         # the index of an element in the original array along dimension that we will flip
-        ik = ((UInt32(ceil(index_in / numelemsinprevdims)) - 1) % numelemsincurrdim) + 1
-        index_out = UInt32(index_in + (numelemsincurrdim - 2ik + 1) * numelemsinprevdims)
+        ik = ((ceil(Int, index_in / numelemsinprevdims) - 1) % numelemsincurrdim) + 1
+
+        index_out = index_in + (numelemsincurrdim - 2ik + 1) * numelemsinprevdims
 
         if 1 ≤ index_in ≤ length(input) && 1 ≤ index_out ≤ length(output)
             @inbounds output[index_out] = input[index_in]
--- a/src/array.jl
+++ b/src/array.jl
@@ -280,6 +280,8 @@ end
 
 ## reversing
 
+# with `start`/`stop` arguments
+
 function _reverse(input::CuVector{T}, output::CuVector{T}) where {T}
     @assert length(input) == length(output)
 
@@ -334,8 +336,9 @@ function Base.reverse(v::CuVector, start=1, stop=length(v))
     return v′
 end
 
-function Base.reverse!(input::CuArray{T, N}, output::CuArray{T, N}; dims::Integer) where {T, N}
+# with `dims` argument
 
+function Base.reverse!(input::CuArray{T, N}, output::CuArray{T, N}; dims::Integer) where {T, N}
     if !(1 ≤ dims ≤ length(size(input)))
       ArgumentError("dimension $dims is not 1 ≤ $dims ≤ $length(size(input))")
     end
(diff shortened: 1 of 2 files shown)


src/array.jl
```diff
@@ -280,56 +280,118 @@ end
...
+        # after reversing by the xth dimension an element at pos
+        # [i1, i2, i3, ... , i{x},            ..., i{n}] will be at
+        # [i1, i2, i3, ... , d{x} - i{x} + 1, ..., i{n}], where
+        # d{x} is the size of the xth dimension
+
+        offset_in = blockDim().x * (blockIdx().x - 1)
+        index_in = offset_in + threadIdx().x
+
+        # the index of an element in the original array along dimension that we will flip
+        ik = ((UInt32(ceil(index_in / numelemsinprevdims)) - 1) % numelemsincurrdim) + 1
```
maleadt: @kraftpunk97 why did you use `UInt32` here? We typically do our indexing in with `Int`s.

kraftpunk97: I thought that it was a safe move to convert indexes to UInt32, so that nothing untoward happened. The C examples that I studied used `unsigned int` indexes. Is that not the case? Sorry, I'm still learning CUDA.

maleadt: No that's fine, in (CUDA) C it's common to use such specific types, but we typically leave those off in Julia unless there's an explicit reason (performance, correctness).

kraftpunk97: @maleadt I'm so sorry I missed this before. The `UInt32` was put there for one more reason. `ceil(index_in / numelemsinprevdims)` returns a `Float64` value, which would make `ik` and `index_out` `Float64` values. This `index_out` is being used as an index values and indices can not be Floats. This is probably the reason why the tests are failing at `@inbounds output[index_out] = input[index_in]`.

Replacing the current  `ceil` call with `ceil(Int, index_in / numelemsinprevdims...` should do the trick.

maleadt: Ah of course, I was too focused on the type not to notice the call to `ceil` :smile:

kraftpunk97:
@maleadt I would like to work on the in-place version of ND-reverse. Do I add to this branch/PR?

maleadt:
> @maleadt I would like to work on the in-place version of ND-reverse. Do I add to this branch/PR?

Sure, but you probably won't be able to push to this branch. Feel free to push to a new one and have me merge, or just open a new PR.

I realized that the ND-kernels should be usable for the 1D case, so I've unified them. The TODO left is to adjust the indexing for the in-place kernel to support multiple dimensions. Note that it works a little differently, it's not one thread that's responsible for reversing a single element, instead independently writing to shared memory and reading from it. It becomes clear with some print statements:

```patch
+++ b/src/array.jl
@@ -307,6 +307,7 @@ function reverse_by_moving(data::CuArray{T, N}, dims::Integer=1) where {T, N}
         if index_in <= length(data)
             index_shared = blockDim().x - threadIdx().x + 1
             @inbounds shared[index_shared] = data[index_in]
+            @cuprintln("thread ", threadIdx().x, ", block ", blockIdx().x, ": shared[", index_shared, "] = data[", index_in, "]")
         end
 
         sync_threads()
@@ -319,6 +320,7 @@ function reverse_by_moving(data::CuArray{T, N}, dims::Integer=1) where {T, N}
         if 1 <= index_out <= length(data)
             index_shared = threadIdx().x
             @inbounds data[index_out] = shared[index_shared]
+            @cuprintln("thread ", threadIdx().x, ", block ", blockIdx().x, ": data[", index_out, "] = shared[", index_shared, "]")
         end
 
         return
```

```
julia> a = cu([1,2,3,4,5,6]); reverse!(a, 2, 5)
6-element CuArray{Float32,1}:
thread 1, block 1: shared[256] = data[1]
thread 2, block 1: shared[255] = data[2]
thread 3, block 1: shared[254] = data[3]
thread 4, block 1: shared[253] = data[4]
 1.0
 5.0
 4.0
 3.0
 2.0
 6.0
```

maleadt:
Ugh, this failure looks to be another case of https://github.com/JuliaGPU/CUDAnative.jl/issues/4: only occurs under `--check-bounds=yes` :sob:

maleadt:
After having confused myself for half a day, this is ready to go. Thanks again, @kraftpunk97!

PR #420, opened 2019-09-17
Wrap syrkx, fix calling errors, add tests

kshyatt:
We were missing wrappers for `syrkx!`. Also, the `xt_syrkx` function was calling the wrong thing. Also, added a bunch of error-throwing tests.


kshyatt:
bors try

kshyatt:
looks like the 1.3 failures are unrelated?

maleadt:
There's a tag of GPUArrays incoming that should resolve 1.3 failures.

PR #422, opened 2019-09-19
Split memory pool robustness improvements

maleadt:


maleadt:
1.3 failure cannot be related. Will need to look into this, but merging for now.

PR #424, opened 2019-09-19
Added in-place ND reverse

kraftpunk97:
For an array of shape `[3, 4, 4, 3, 6, 5]`,

```
julia> @btime reverse_by_moving(v_d, 3)
  31.330 μs (26 allocations: 1.11 KiB)
```
cc @maleadt

### DIFF
--- a/src/array.jl
+++ b/src/array.jl
@@ -297,32 +297,37 @@ function reverse_by_moving(data::CuArray{T, N}, dims::Integer=1) where {T, N}
     function kernel(data::CuDeviceArray{T, N}) where {T, N}
         #@cuprintf("%ld  %ld %ld\n", threadIdx().x, blockIdx().x, blockDim().x)
 
-        #shared = @cuDynamicSharedMem(T, blockDim().x)
+        shared = @cuDynamicSharedMem(T, blockDim().x)
 
-        # Each thread is responsible for swapping an element with another element
-        # that is currently in its destination position.
+        # Each thread is responsible for shifting one element.
 
-        # Calculate the index of the element to swapped.
+        # Copy the block as is into shared memory.
         offset_in = blockDim().x * (blockIdx().x - 1)
         index_in  = offset_in + threadIdx().x
 
-        # Calculate its destination...
+        if index_in ≤ length(data)
+            index_shared = threadIdx().x
+            @inbounds shared[index_shared] = data[index_in]
+        end
+
+        sync_threads()
+
+        # Calculate the index of the new position.
         ik = ((ceil(Int, index_in / numelemsinprevdims) - 1) % numelemsincurrdim) + 1
         index_out = index_in + (numelemsincurrdim - 2ik + 1) * numelemsinprevdims
 
-        # And hit swap!
-        @inbounds temp = data[index_in]
-        @inbounds data[index_in] = data[index_out]
-        @inbounds data[index_out] = temp
+        # Move the values...
+        if 1 ≤ index_out ≤ length(data)
+            @inbounds data[index_out] = shared[index_shared]
+        end
 
         return nothing
     end
 
     nthreads = 256
     nblocks = ceil(Int, length(data) / nthreads)
-    #shmem = nthreads * sizeof(T)
-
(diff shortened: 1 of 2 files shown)


kraftpunk97:
Okay, so turns out that there is a way to implement in-place reverse without \*\*technically\*\* using  shared memory. I just had to calculate the index of the destination position and swap the original element with the element at that position. For a vector of size 2^15, the shared memory implementation takes about 27 μs, while this implementation uses 30 μs.

maleadt:
The reason you do it the other way is that you get nicely coalesced global memory access patterns -- the only thing you can really optimize in a kernel like this. I'll have a go at some more detailed benchmarks when I have the time.

kraftpunk97:
Oh I see. I'll make the changes.

maleadt:
Might not be as relevant on more modern GPUs (which are a little smarter wrt. coalescing), so if it isn't necessary we should go for a single, simple kernel :-)

kraftpunk97:
Ok, so I reimplemented the function using shared memory. Is this better?

maleadt:
Yes, but it's still not coalescing because you're accessing the global data in reverse mode / storing in the shared memory buffer in forwards mode. The idea with coalesced accesses is that consecutive threads access memory contiguously and ordered, for the memory controller to merge. That's why the original shared memory implementation reads global memory forwards, stores it backwards in the shared memory buffer, and then switches direction reading out the shared memory buffer forwards and storing into global memory forwards again.

kraftpunk97:
Okay. I'll have to look into memory coalescing techniques. I'll study them and then try to reimplement this function.

maleadt:
The idea is basically to reverse the first part of your kernel, i.e. to write in `index_shared = blockDim().x - threadIdx().x` and reverse your source index calculation.

kraftpunk97:
@maleadt Sorry for the late response. I understand you're asking me to store the elements in shared memory in reverse order, and copy them back in forward order, but, in my case, the destination address of an element is dependent on `index_in`, unlike your implementation, which calculates `index_out` directly from `blockIdx().x` and `threadIdx().x`, without calculating `index_in` first. From what I've gathered from articles on memory coalescing (and I may be completely wrong about this), MC is an issue when threads of the same block try to access strided global memory locations,  but as far as I can tell, that is not an issue here: threads are reading elements from contiguous memory locations, saving to a shared memory cache, and then rewriting to almost contiguous memory locations. So I think we have handled the issue of MC. What am I missing?

Also how does reversing the order the when writing to shared memory and copying to global memory in forward order help with MC in `_reverse` in `array.jl`? Again, you're just reading and rewriting to contiguous memory locations. The order of the read and writes shouldn't matter.

maleadt:
> in my case, the destination address of an element is dependent on `index_in`, unlike your implementation, which calculates `index_out` directly from `blockIdx().x` and `threadIdx().x`, without calculating `index_in` first

Well, yeah, the index calculation needs to be decoupled.

> MC is an issue when threads of the same block try to access strided global memory locations

No, MC improves memory when accessing subsequent memory addresses by consecutive threads. https://devblogs.nvidia.com/using-shared-memory-cuda-cc/ illustrates with an array reversal example:

> The reason shared memory is used in this example is to facilitate global memory coalescing on older CUDA devices (Compute Capability 1.1 or earlier). Optimal global memory coalescing is achieved for both reads and writes because global memory is always accessed through the linear, aligned index t. The reversed index tr is only used to access shared memory, which does not have the sequential access restrictions of global memory for optimal performance. The only performance issue with shared memory is bank conflicts, which we will discuss later. (Note that on devices of Compute Capability 1.2 or later, the memory system can fully coalesce even the reversed index stores to global memory. But this technique is still useful for other access patterns, as I’ll show in the next post.)

> The order of the read and writes shouldn't matter.

It does, as per the snippet above on older hardware.

Anyway, I didn't want this to drag along as long as it did, sorry for that. I've had a go at doing the suggested changes in https://github.com/JuliaGPU/CuArrays.jl/pull/416.

kraftpunk97:
That is alright. Closing PR now...

PR #430, opened 2019-09-26
Simplify CI and commit Manifest

maleadt:
This PR adds Manifests and changes to a new CI template that uses them in order to have more reproducible CI, and make it easier to depend on master branches of dependencies (e.g. CUDAnative).

CI currently failing because I had pushed a broken `.yml` and at some point that file is being cached. Let's retry tomorrow.

cc @dhairyagandhi96

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,5 @@
 *.jl.*.cov
 *.jl.mem
 deps/ext.jl
-Manifest.toml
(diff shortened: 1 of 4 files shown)


dhairyagandhi96:
Cool! Thanks! We should switch to the new templates in Flux also.

PR #434, opened 2019-09-27
Use CuArrays.zeros instead of broadcasting.

maleadt:


maleadt:
All-green CI (well, only Flux fails, but looking at that as well). That's been a while!

PR #439, opened 2019-10-01
remove extraneous where and add some tests

kshyatt:


kshyatt:
bors try

maleadt:
No need to queue another test if you're part of JuliaGPU, hence the timeout on the second one (I reduced amount of parallel builds due to a couple of GPU OOM situations).

kshyatt:
Oh dang, I had no idea. Well then, good to merge?

Issue #456, opened 2019-10-22
Slow view construction


maleadt:
```
julia> using BenchmarkTools

julia> N = 512
512

julia> A = rand(N,N,N);

julia> @benchmark view($A, :, :, 20)
BenchmarkTools.Trial: 
  memory estimate:  64 bytes
  allocs estimate:  1
  --------------
  minimum time:     7.355 ns (0.00% GC)
  median time:      20.944 ns (0.00% GC)
  mean time:        26.865 ns (16.39% GC)
  maximum time:     44.034 μs (99.96% GC)
  --------------
  samples:          10000
  evals/sample:     1000

julia> using CuArrays

julia> B = CuArray(A);

julia> typeof(view(B, :, :, 20))
CuArray{Float64,2}

julia> @benchmark view($B, :, :, 20)
BenchmarkTools.Trial: 
  memory estimate:  112 bytes
  allocs estimate:  3
  --------------
  minimum time:     153.789 ns (0.00% GC)
  median time:      188.939 ns (0.00% GC)
  mean time:        674.696 ns (8.76% GC)
  maximum time:     4.315 ms (12.43% GC)
  --------------
  samples:          9078
  evals/sample:     815
```

This matters especially for batched interfaces:

```
julia> @benchmark [view($A, :, :, i) for i in 1:$N]
BenchmarkTools.Trial: 
  memory estimate:  36.17 KiB
  allocs estimate:  515
  --------------
  minimum time:     4.094 μs (0.00% GC)
  median time:      11.581 μs (0.00% GC)
  mean time:        14.128 μs (17.04% GC)
  maximum time:     6.267 ms (96.64% GC)
  --------------
  samples:          10000
  evals/sample:     7

julia> @benchmark [view($B, :, :, i) for i in 1:$N]
BenchmarkTools.Trial: 
  memory estimate:  60.17 KiB
  allocs estimate:  1539
  --------------
  minimum time:     79.919 μs (0.00% GC)
  median time:      96.215 μs (0.00% GC)
  mean time:        423.850 μs (9.58% GC)
  maximum time:     3.232 s (12.56% GC)
  --------------
  samples:          10000
  evals/sample:     1
```

Especially bad since it essentially boils down (for the batched interfaces at least) to:

```
ulia> using CUDAdrv: CuPtr

julia> stride = N*N
262144

julia> T = eltype(A)
Float64

[...]

PR #457, opened 2019-10-23
Adapt to CUDAdrv.jl#167

maleadt:


maleadt:
FWIW, these large-scale refactors aren't just for fun: the overhead of calling `getrf_batched!` goes down from a whopping 5ms to about 800us. Still way too much, and https://github.com/JuliaGPU/CuArrays.jl/issues/456 is part of the reason why, but getting rid of the inherent type instability of `CUDAdrv.Mem.Buffer` as well as the global refcounting state is a nice improvement already.

Issue #440, opened 2019-10-02
Matrix inversion for CuArray


xukai92:
I attempted to use a wrap over CuArrays.CUBLAS.matinv_batched, i.e. Base.inv(x::CuArray{<:Real,2}) = CuArrays.CUBLAS.matinv_batched([x])[2][1]. But this only supports matrix smaller than 32x32.

PR #459, opened 2019-10-24
Implement matrix division using CUSOLVER

xukai92:
Towards addressing https://github.com/JuliaGPU/CuArrays.jl/issues/440 via matrix division as the first step (https://github.com/JuliaGPU/CuArrays.jl/issues/440#issuecomment-537942479)


src/CuArrays.jl
```diff
@@ -62,6 +62,7 @@ include("utils.jl")
 include("indexing.jl")
 include("broadcast.jl")
 include("matmul.jl")
+include("matinv.jl")
```
maleadt: `matinv.jl` is pretty specific, maybe this functionality should go in `linalg.jl`, together with `matmul.jl`?

Manifest.toml
```diff
@@ -74,9 +74,9 @@ version = "4.0.0"
 
 [[DataStructures]]
 deps = ["InteractiveUtils", "OrderedCollections"]
```
maleadt: if the manifest changes aren't needed, please leave them out

test/runtests.jl
```diff
@@ -18,16 +18,16 @@ allowscalar(false)
 
 @testset "CuArrays" begin
 
-include("base.jl")
```
maleadt: accidentally committed, I guess.

src/CuArrays.jl
xukai92: I thought `linalg.jl` is for overloading `LinearAlgebra` module. There is also a `matmul.jl` already existing. But I'm happy to move `matinv.jl` into `linalg.jl` - no problem.

Manifest.toml
xukai92: I was not able to run tests without `] update` - no sure why. I can revert the change though.

maleadt: It is, but in Base `\(::Matrix, ::Matrix)` is implemented in the LinearAlgebra stdlib too, so I guess this makes sens?

src/CuArrays.jl
maleadt: It is, but in Base \(::Matrix, ::Matrix) is implemented in the LinearAlgebra stdlib too, so I guess this makes sens?

Manifest.toml
xukai92: Oh I didn't release this! Let me move the code into `linalg.jl` then!

maleadt: You were using `julia --project` right? The manifest should always be valid, as it's what the CI infrastructure uses. Try getting rid of it, and we'll see what CI thinks about it :slightly_smiling_face:

src/CuArrays.jl
xukai92: Oh I didn't release this! Let me move the code into `linalg.jl` then!

Manifest.toml
xukai92: I did with `--project=@.` inside the folder. I got errors about `CUstream` not defined.

maleadt: You probably didn't `instantiate` then (to have Pkg actually check out the versions in the Manifest).

xukai92: Oh I see. Thanks! I will revert this then.

xukai92:
Changes done.

maleadt:
bors try

PR #460, opened 2019-10-25
Improve type constraint for `\`

xukai92:


src/linalg.jl
```diff
@@ -49,9 +49,7 @@ end
 
 # Matrix division
 
-const CuOrAdj = Union{CuVecOrMat, LinearAlgebra.Adjoint{T, CuVecOrMat{T}}, LinearAlgebra.Transpose{T, CuVecOrMat{T}}} where {T<:AbstractFloat}
-
-function Base.:\(_A::AT1, _B::AT2) where {AT1<:CuOrAdj, AT2<:CuOrAdj}
+function Base.:\(_A::CuMatOrAdj, _B::CuOrAdj)
```
xukai92: A has to be a matrix while B could be a vector or matrix.

src/array.jl
```diff
@@ -20,6 +20,9 @@ end
 CuVector{T} = CuArray{T,1}
 CuMatrix{T} = CuArray{T,2}
 CuVecOrMat{T} = Union{CuVector{T},CuMatrix{T}}
+CuMatOrAdj{T} = Union{CuMatrix, LinearAlgebra.Adjoint{T, CuMatrix{T}}, LinearAlgebra.Transpose{T, CuMatrix{T}}}
+CuOrAdj{T} = Union{CuVecOrMat, LinearAlgebra.Adjoint{T, CuVecOrMat{T}}, LinearAlgebra.Transpose{T, CuVecOrMat{T}}}
```
xukai92: Are you happy with this?

xukai92:
Actually fixed a bug for type constraint in https://github.com/JuliaGPU/CuArrays.jl/pull/460/commits/3eb88d5c31fe930c525e532c23d63fcb30e16648

src/array.jl
maleadt: Better add these type aliases to the linalg file

xukai92: Moved.

maleadt:
bors try

Issue #409, opened 2019-09-06
Mem.Buffer's can be returned to poll even though they didn't originate from it


vchuravy:
```julia
using CuArrays
using CUDAdrv
import CUDAdrv.Mem

buf = Mem.alloc(Mem.HostBuffer, 1024*sizeof(Float64))
arr = CuArray{Float64,1}(buf, (1024,))

julia> all(isempty, CuArrays.pools_avail)
true
julia> CuArrays.unsafe_free!(arr)
julia> all(isempty, CuArrays.pools_avail)
false
julia> filter(!isempty, CuArrays.pools_avail)
1-element Array{Array{CUDAdrv.Mem.Buffer,1},1}:
 [HostBuffer(Ptr{Nothing} @0x00007f745fa0a000, 8192, CuContext(Ptr{Nothing} @0x0000000003466190, false, true), false)]
```

cc: @lcw

maleadt:
Couple of comments, after the recent refactor.

1) CuArray's now container ordinary CuPtr's
2) You should always `convert` a HostBuffer to an appropriate pointer (Ptr or CuPtr)
3) To convert a HostBuffer to a CuPtr, it needs the Mem.HOSTALLOC_DEVICEMAP flag
4) You should use `unsafe_wrap(CuArray)` to create an array from an arbitrary `CuPtr`. The `own` flag to `unsafe_wrap` indicates whether Julia should call `cuMemFree` as part of finalization, in your case you'd need a custom `finalizer` to call `cuMemHostFree`.
5) `unsafe_free!` serves to put objects back into the pool, so it won't do anything on `unsafe_wrap`ped objects (regardless of `own`).

```julia
julia> buf = Mem.alloc(Mem.HostBuffer, 1024*sizeof(Float64), Mem.HOSTALLOC_DEVICEMAP)
CUDAdrv.Mem.HostBuffer(Ptr{Nothing} @0x00007fbf8cc02000, 8192, CuContext(Ptr{Nothing} @0x0000000002e138d0, false, true), true)

julia> cpu_ptr = convert(Ptr{Float64}, buf)
Ptr{Float64} @0x00007fbf8cc02000

julia> gpu_ptr = convert(CuPtr{Float64}, buf)
CuPtr{Float64}(0x00007fbf8cc02000)

julia> arr = unsafe_wrap(CuArray{Float64}, gpu_ptr, (1024,); own=false);

julia> CuArrays.unsafe_free!(arr)

[...]

PR #467, opened 2019-10-31
implement CuIterator for batching arrays to the GPU

jrevels:
I'm hitting what I believe is likely a common use case for CuArrays. My workload is essentially a map-reduce of the form:

```julia
some_reduction(f(batch::Tuple{Vararg{Array}})::Number for batch in batches)
```

Assuming `f` plays nicely with CuArrays, one way to get this running on a GPU is:

```julia
cubatches = (map(x -> adapt(CuArray, x), batch) for batch in batches)
some_reduction(f(cubatch) for cubatch in cubatches)
```

Unfortunately, this approach is a poor one when `batches` doesn't entirely fit in GPU memory. As the caller, I can assert that the old iterations' batches don't need to be kept around, so ideally, I'd have a mechanism to leverage to simply reuse old iterations' memory instead of allocating more. This PR implements such a mechanism: a `CuIterator` that maintains a memory pool and exploits the assumption that previous iterations' memory can be reused.

This is just a rough POC sketch right now just to express what I'm trying to get at...it's probably dumb in ways I can't yet comprehend. AFAICT I'd like to do the following before merging:

- [x] instead of the current shape/eltype-matching approach, just allocate to a device buffer (~`CUDAdrv.UnifiedBuffer`? I have no idea what I'm doing 😅 EDIT: oh is this how CUDAdrv exposes UVM or is that separate?~), increasing the buffer size as necessary once larger batches are encountered
- [x] ~leverage `CUDAdrv.prefetch` to asynchronously move values to GPU memory so as to not artificially block the iterator's consumer~ EDIT: I'm not sure this is necessary anymore assuming the `copyto!` call we're using is asynchronous?
- [x] make sure to free the pool once iteration is finished
- [x] docs
- [x] tests

In a future PR, we could add a feature for `CuIterator` to utilize UVM if the caller is in an environment where that's supported.

cc @vchuravy (thanks for discussing this with me earlier!)

### DIFF
--- a/src/iterator.jl
+++ b/src/iterator.jl
@@ -1,28 +1,38 @@
-struct CuIterator{B,P}
+struct CuIterator{B}
     batches::B
-    pool::Vector{P}
-    pool_limit::Int
+    initial_pool_size::Ref{Int}
+    pool::Ref{Mem.DeviceBuffer}
 end
 
-CuIterator(batches, pool=Any[]) = CuIterator(batches, pool, 10)
-
-pool_key(x) = eltype(x) => size(x)
+function CuIterator(batches, initial_pool_size::Integer=0)
+    return CuIterator(batches, Ref(initial_pool_size), Ref{Mem.DeviceBuffer}())
+end
 
 function Base.iterate(c::CuIterator, state...)
     item = iterate(c.batches, state...)
     if item === nothing
-        foreach(batch -> foreach(unsafe_free!, batch), c.pool)
-        empty!(c.pool)
+        Mem.free(c.pool[])
         return nothing
     end
     batch, next_state = item
-    i = findfirst(allocated -> pool_key.(allocated) == pool_key.(batch), c.pool)
-    if i === nothing
-        cubatch = map(x -> adapt(CuArray, x), batch)
-        push!(c.pool, cubatch)
-    else
-        cubatch = map(copyto!, c.pool[i], batch)
+    required_pool_size = sum(sizeof, batch)
+    if isempty(state)
+        c.initial_pool_size[] = max(required_pool_size, c.initial_pool_size[])
+        c.pool[] = Mem.alloc(Mem.DeviceBuffer, c.initial_pool_size[])
+    elseif required_pool_size > sizeof(c.pool[])
+        Mem.free(c.pool[])
+        c.initial_pool_size[] = required_pool_size
+        c.pool[] = Mem.alloc(Mem.DeviceBuffer, required_pool_size)
     end
-    length(c.pool) > c.pool_limit && foreach(unsafe_free!, popfirst!(c.pool))
+    pool = c.pool[]
+    offset = 0
+    cubatch = map(batch) do array
+        @assert array isa AbstractArray
+        ptr = Base.unsafe_convert(CuPtr{eltype(array)}, pool.ptr + offset)
+        cuarray = unsafe_wrap(CuArray, ptr, size(array); own=false);
+        copyto!(cuarray, array)
+        offset += sizeof(array)
+        return cuarray
+    end
     return cubatch, next_state
 end
--- a/src/iterator.jl
+++ b/src/iterator.jl
@@ -1,30 +1,29 @@
-struct CuIterator{B}
+mutable struct CuIterator{B}
     batches::B
-    initial_pool_size::Ref{Int}
-    pool::Ref{Mem.DeviceBuffer}
-end
-
-function CuIterator(batches, initial_pool_size::Integer=0)
-    return CuIterator(batches, Ref(initial_pool_size), Ref{Mem.DeviceBuffer}())
+    initial_pool_size::Int
+    pool::Mem.DeviceBuffer
+    function CuIterator(batches, initial_pool_size=0)
+        return new{typeof(batches)}(batches, initial_pool_size)
+    end
 end
 
 function Base.iterate(c::CuIterator, state...)
     item = iterate(c.batches, state...)
     if item === nothing
-        Mem.free(c.pool[])
+        isdefined(c, :pool) && Mem.free(c.pool)
         return nothing
     end
     batch, next_state = item
     required_pool_size = sum(sizeof, batch)
     if isempty(state)
-        c.initial_pool_size[] = max(required_pool_size, c.initial_pool_size[])
-        c.pool[] = Mem.alloc(Mem.DeviceBuffer, c.initial_pool_size[])
-    elseif required_pool_size > sizeof(c.pool[])
-        Mem.free(c.pool[])
-        c.initial_pool_size[] = required_pool_size
-        c.pool[] = Mem.alloc(Mem.DeviceBuffer, required_pool_size)
+        c.initial_pool_size = max(required_pool_size, c.initial_pool_size)
+        c.pool = Mem.alloc(Mem.DeviceBuffer, c.initial_pool_size)
+    elseif required_pool_size > sizeof(c.pool)
+        Mem.free(c.pool)
+        c.initial_pool_size = required_pool_size
+        c.pool = Mem.alloc(Mem.DeviceBuffer, required_pool_size)
     end
-    pool = c.pool[]
+    pool = c.pool
     offset = 0
     cubatch = map(batch) do array
         @assert array isa AbstractArray
--- a/src/iterator.jl
+++ b/src/iterator.jl
@@ -1,37 +1,15 @@
 mutable struct CuIterator{B}
     batches::B
-    initial_pool_size::Int
-    pool::Mem.DeviceBuffer
-    function CuIterator(batches, initial_pool_size=0)
-        return new{typeof(batches)}(batches, initial_pool_size)
-    end
+    previous::Any
+    CuIterator(batches) = new{typeof(batches)}(batches)
 end
 
 function Base.iterate(c::CuIterator, state...)
     item = iterate(c.batches, state...)
-    if item === nothing
-        isdefined(c, :pool) && Mem.free(c.pool)
-        return nothing
-    end
+    isdefined(c, :previous) && foreach(unsafe_free!, c.previous)
+    item === nothing && return nothing
     batch, next_state = item
-    required_pool_size = sum(sizeof, batch)
-    if isempty(state)
-        c.initial_pool_size = max(required_pool_size, c.initial_pool_size)
-        c.pool = Mem.alloc(Mem.DeviceBuffer, c.initial_pool_size)
-    elseif required_pool_size > sizeof(c.pool)
-        Mem.free(c.pool)
-        c.initial_pool_size = required_pool_size
-        c.pool = Mem.alloc(Mem.DeviceBuffer, required_pool_size)
-    end
-    pool = c.pool
-    offset = 0
-    cubatch = map(batch) do array
-        @assert array isa AbstractArray
-        ptr = Base.unsafe_convert(CuPtr{eltype(array)}, pool.ptr + offset)
-        cuarray = unsafe_wrap(CuArray, ptr, size(array); own=false);
-        copyto!(cuarray, array)
-        offset += sizeof(array)
-        return cuarray
-    end
+    cubatch = map(x -> adapt(CuArray, x), batch)
+    c.previous = cubatch
     return cubatch, next_state
 end
--- a/src/iterator.jl
+++ b/src/iterator.jl
@@ -10,7 +10,11 @@ pool_key(x) = eltype(x) => size(x)
 
 function Base.iterate(c::CuIterator, state...)
     item = iterate(c.batches, state...)
-    item === nothing && return nothing
+    if item === nothing
+        map(batch -> map(unsafe_free!, batch), c.pool)
+        empty!(c.pool)
+        return nothing
+    end
     batch, next_state = item
     i = findfirst(allocated -> pool_key.(allocated) == pool_key.(batch), c.pool)
     if i === nothing
--- a/src/iterator.jl
+++ b/src/iterator.jl
@@ -23,6 +23,6 @@ function Base.iterate(c::CuIterator, state...)
     else
         cubatch = map(copyto!, c.pool[i], batch)
     end
-    length(c.pool) > c.pool_limit && map(unsafe_free!, popfirst!(c.pool))
+    length(c.pool) > c.pool_limit && foreach(unsafe_free!, popfirst!(c.pool))
     return cubatch, next_state
 end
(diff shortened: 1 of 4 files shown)


vchuravy:
Also see Tim's comment here https://github.com/JuliaGPU/CuArrays.jl/issues/409#issuecomment-546327373

src/iterator.jl
```diff
@@ -0,0 +1,28 @@
...
+    pool_limit::Int
+end
+
+CuIterator(batches, pool=Any[]) = CuIterator(batches, pool, 10)
+
+pool_key(x) = eltype(x) => size(x)
+
+function Base.iterate(c::CuIterator, state...)
+    item = iterate(c.batches, state...)
+    if item === nothing
```
jrevels: Note that if we take this kind of approach for freeing the pool, callers which break out of iteration early won't get the nice automatic freeing behavior

src/iterator.jl
```diff
@@ -0,0 +1,28 @@
...
+    end
+    batch, next_state = item
+    i = findfirst(allocated -> pool_key.(allocated) == pool_key.(batch), c.pool)
+    if i === nothing
+        cubatch = map(x -> adapt(CuArray, x), batch)
+        push!(c.pool, cubatch)
+    else
+        cubatch = map(copyto!, c.pool[i], batch)
+    end
+    length(c.pool) > c.pool_limit && map(unsafe_free!, popfirst!(c.pool))
```
ararslan: Could do
```suggestion
    length(c.pool) > c.pool_limit && foreach(unsafe_free!, popfirst!(c.pool))
```
since you aren't using the result of `map`

jrevels:
Okay, just paired up with @vchuravy to get this thing using a `DeviceBuffer` for the pool instead of the silly approach I had before. From some toy benchmarks seems like things are kinda working :D

Now that something basic is in place, I'd love to get @vchuravy's / @maleadt's thoughts both on this current implementation and potential improvements e.g. are there opportunities for prefetching/asynchronicity that I'm missing/should be thinking about?

src/iterator.jl
```diff
@@ -0,0 +1,38 @@
+struct CuIterator{B}
+    batches::B
+    initial_pool_size::Ref{Int}
+    pool::Ref{Mem.DeviceBuffer}
```
maleadt: Why an immutable struct with `Ref` and not just a mutable struct?

src/iterator.jl
```diff
@@ -0,0 +1,38 @@
...
+    end
+    batch, next_state = item
+    required_pool_size = sum(sizeof, batch)
+    if isempty(state)
+        c.initial_pool_size[] = max(required_pool_size, c.initial_pool_size[])
+        c.pool[] = Mem.alloc(Mem.DeviceBuffer, c.initial_pool_size[])
+    elseif required_pool_size > sizeof(c.pool[])
+        Mem.free(c.pool[])
+        c.initial_pool_size[] = required_pool_size
+        c.pool[] = Mem.alloc(Mem.DeviceBuffer, required_pool_size)
```
maleadt: Does not free the old pool?

src/iterator.jl
```diff
@@ -0,0 +1,38 @@
...
+    end
+    pool = c.pool[]
+    offset = 0
+    cubatch = map(batch) do array
+        @assert array isa AbstractArray
+        ptr = Base.unsafe_convert(CuPtr{eltype(array)}, pool.ptr + offset)
+        cuarray = unsafe_wrap(CuArray, ptr, size(array); own=false);
+        copyto!(cuarray, array)
```
maleadt: That would only work with plain arrays. What if the array was, e.g., an `Adjoint`? You'd need to use Adapt for that again.

src/iterator.jl
jrevels: Didn't want anybody mutating `batches` on me, but that's probably overly defensive on my part; will change it

src/iterator.jl
jrevels: Ah, I thought the `Mem.free(c.pool[])` above would?

jrevels:
Thanks for taking a look, Tim!

> Do you have a motivating example?

I'm training a Flux model, and my batches are too numerous to all simultaneously fit in memory. Flux's `train!` function expects batch iterators as inputs, so `CuIterator` is my attempt at meeting that expectation. I could just write my own training loop, but this seemed like an opportunity to start messing about with JuliaGPU stuff 😃If there's a better existing pattern I can use for this use case, though, that would also be dope!

> CuArrays already provides a memory pool, you could just create new CuArray objects and CuArrays.unsafe_free! the results of previous iterations to make them available for reuse. That would also make it possible to use adapt again, and support a wider range of arrays.

IIUC this is similar to what I actually implemented in this [PR's original rough sketch](https://github.com/JuliaGPU/CuArrays.jl/blob/fd792274057b566bf2ce8f6f595336f973b48a79/src/iterator.jl), but for some reason I felt like that approach was hacky for the case where `batches`'s iterations produced arrays of varying size. After discussing with Valentin, it seemed like the better approach would be to grow a pool to the maximum required size, and just reuse that. @vchuravy might have had other tradeoffs in mind as well.

I also thought one thing that would be cool would be to let the caller basically over-allocate the pool up front (using `initial_pool_size`) and then asynchronously prefetch future iterations into the pool until it's full. This would be useful in my case as well, where each individual batch isn't that big - I just have a lot of them, and I know how much memory I can spend up front and still leave enough room left for the computation I'm doing. Basically using `initial_pool_size` as a tuning parameter for throughput. AFAIK this isn't what's actually implemented here (nor do I know enough yet to implement it, e.g. I'm still a noob wrt figuring out which CUDAdrv/CuArray calls are blocking), just an idea. 

Seems like from discussions with Valentin, UVM would get us a lot of this stuff for free if it were more ubiquitous. Maybe I'd also get this kind of nice behavior for free though if this PR did a better job at reusing CuArrays' existing memory pool? Or perhaps it's not a good idea anyway? Down to change the approach here however makes sense, just let me know 🙂

jrevels:
After talking with @SimonDanisch - one thing I could do is literally just keep the pointer for the previous iteration around and call `unsafe_free!` on it. Would that be reusing CuArrays' existing memory pool in the right way here?

```julia
using CuArrays: CuArray, unsafe_free!, adapt

mutable struct SimpleCuIterator{B}
    batches::B
    previous::Any
    SimpleCuIterator(batches) = new{typeof(batches)}(batches)
end

function Base.iterate(c::SimpleCuIterator, state...)
    item = iterate(c.batches, state...)
    isdefined(c, :previous) && foreach(unsafe_free!, c.previous)
    item === nothing && return nothing
    batch, next_state = item
    cubatch = map(x -> adapt(CuArray, x), batch)
    c.previous = cubatch
    return cubatch, next_state
end
```

Seems like this works nearly as well as the more complicated implementation here, but trying to design good MWE benchmarks has made me reconsider whether this is actually useful/my workload actually requires it...it seemed to relieve memory pressure in my actual application, but that situation is more complicated and the improvements I originally attributed to this could've been conflated with other changes.

CuArrays is just polling the Julia GC occasionally to figure out which pool-allocated objects can be freed, right?  Then the potential advantage of `CuIterator` versus e.g. `(map(x -> adapt(CuArray, x), batch) for batch in batches)` should be that it can mark objects as "freeable" earlier/more often than the polling rate would otherwise allow. Thus, AFAICT, a workload that should benefit from `CuIterator` is one where each iteration takes way less time than the polling rate and allocates enough that waiting for the next poll to free objects would result in throttling. However, I'm having trouble coming up with a naive MWE benchmark that demonstrates an actual improvement here...is my understanding just incorrect? 

Maybe this was all just an educational exercise 😛if so, at least it has indeed taught me a lot. If anybody's curious, here's [the benchmark gist](https://gist.github.com/jrevels/26c760131c80662771d36b5965239b85) I've been playing around with.

src/iterator.jl
maleadt: Oh oops, I must be blind.

maleadt:
Sorry for not being more responsive here. I'll have a look at your benchmarks as soon as I have the time (and have worked through the onslaught of package loading issues) :slightly_smiling_face: 

> Then the potential advantage of `CuIterator` versus e.g. `(map(x -> adapt(CuArray, x), batch) for batch in batches)` should be that it can mark objects as "freeable" earlier/more often than the polling rate would otherwise allow.

Yeah, there's definitely advantages of the iterator-based approach where you have more information to early release buffers. What I meant with "manual memory management" is you calling into CUDAdrv to allocate objects; bypassing CuArrays like that has a couple of issues (for one, if CuArrays is caching lots of freed buffers for later reuse, your calls into `CUDAdrv.Mem.alloc` might OOM, whereas calling into the CuArrays pool would reuse or force release of those cached buffers).

vchuravy:
I think the important bit here is that you can actually over-allocate (for the largest batch size) and re-use that buffer repeatedly.

The worst case benchmark is is something like:

```
((zeros(N, N), ones(N, N)) for N in (2^i for i in 6:10) )
```

as input to the iterator. I agree with Tim's concerns that the current approach isn't as composable as it should be since it doesn't handle
`Adjoint` et. al. well, and only works on plain old arrays. Benchmarks metrics should probably be: Memory usage (and stability) + Latency the big win will be to reliable re-use an/the same allocation.

jrevels:
> ...What I meant with "manual memory management" is you calling into CUDAdrv to allocate objects...

> ...doesn't handle Adjoint et. al. well, and only works on plain old arrays...

But that's not true of the [most recent iteration of this](https://github.com/JuliaGPU/CuArrays.jl/pull/467#issuecomment-551917668) posted before these comments, right? Just wanted to check my understanding before switching out this branch's implementation again.

> I think the important bit here is that you can actually over-allocate (for the largest batch size) and re-use that buffer repeatedly.

I think the over-allocation is only really useful (at least in my workload) if you can asynchronously prefetch iterations into "leftover" space in the buffer. The simpler implementation posted above doesn't implement this either way, though. 

I wonder whether the right approach here requires using one of CuArrays' memory pool implementations directly (I should be looking in the `src/memory` directory, right?)

maleadt:
> But that's not true of the [most recent iteration of this](https://github.com/JuliaGPU/CuArrays.jl/pull/467#issuecomment-551917668) posted before these comments, right? Just wanted to check my understanding before switching out this branch's implementation again.

Yes, sorry, I was just clarifying that older comment.

jrevels:
If this is desirable in its current form, I can block out some time over the following weeks to rebase + add docs/tests + file issues for all the follow-up work discussed here and get this PR out of draft mode.

maleadt:
> If this is desirable in its current form, I can block out some time over the following weeks to rebase + add docs/tests + file issues for all the follow-up work discussed here and get this PR out of draft mode.

Definitely! Seems very useful functionality. I did take a look at the benchmarks, as I was mainly interested in whether the iterator with manual memory management would outperform our memory pool, but it seems like most allocations happen outside of the iterator so the performance gains are expected to be minimal. But if your applications show otherwise, I'm happy to merge the original `CuIterator` too (although it should probably call `CuArrays.alloc` and not `CUDAdrv.Mem.alloc`).

jrevels:
Finally got around to adding basic docs/tests for this; we've been using this little utility at Beacon for awhile now so will be great to switch to an open-source version :D I still think the async iteration prefetch thing could be a nice feature for this thing, but can leave to follow-up work.

The test is pretty basic, let me know if there's any improvements I should make. Also let me know if there's any reimplementation/updates work to do w.r.t. new developments on master since I last touched this PR.

Is there a place where I should reference the `CuIterator` docstring in wider documentation? Seems like API docs are a WIP here and that general usage docs are in CUDA.jl?

jrevels:
friendly bump :) let me know if there's anything I should do here

maleadt:
Sorry! This looks good to go, so let's see what CI thinks: 
bors try



> Is there a place where I should reference the `CuIterator` docstring in wider documentation? Seems like API docs are a WIP here and that general usage docs are in CUDA.jl?

Correct, so we should think about demoing this over at CUDA.jl; I think it could be a useful addition for a section on memory considerations (design of the pool, how to print stats, unsafe_free, use this iterator, etc).

jrevels:
Rebased against master to resolve conflicts. Not sure I have permissions to do this but:

bors try

> I think it could be a useful addition for a section on memory considerations (design of the pool, how to print stats, unsafe_free, use this iterator, etc).

Yeah, that sounds like it'd be really helpful in general!

maleadt:
You're triggering scalar iteration when comparing the GPU with CPU batches:

```
 CuIterator: Error During Test at /builds/JuliaGPU/CuArrays.jl/test/iterator.jl:12
  Test threw exception
  Expression: batch == cubatch
  scalar getindex is disallowed
  Stacktrace:
   [1] error(::String) at ./error.jl:33
   [2] assertscalar(::String) at /builds/JuliaGPU/CuArrays.jl/.julia/packages/GPUArrays/JtsHd/src/host/indexing.jl:41
   [3] getindex at /builds/JuliaGPU/CuArrays.jl/.julia/packages/GPUArrays/JtsHd/src/host/indexing.jl:86 [inlined]
   [4] iterate at ./abstractarray.jl:914 [inlined]
   [5] iterate at ./abstractarray.jl:912 [inlined]
   [6] _zip_iterate_some at ./iterators.jl:350 [inlined]
   [7] _zip_iterate_some at ./iterators.jl:352 [inlined]
   [8] _zip_iterate_all at ./iterators.jl:342 [inlined]
   [9] iterate at ./iterators.jl:332 [inlined]
   [10] ==(::Array{Float32,3}, ::CuArray{Float32,3,Nothing}) at ./abstractarray.jl:1775
   [11] ==(::Array{Array{Float32,3},0}, ::Array{CuArray{Float32,3,Nothing},0}) at ./abstractarray.jl:1776
   [12] eval_test(::Expr, ::Expr, ::LineNumberNode, ::Bool) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.3/Test/src/Test.jl:245
   [13] top-level scope at /builds/JuliaGPU/CuArrays.jl/test/iterator.jl:12
   [14] top-level scope at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.3/Test/src/Test.jl:1107
   [15] top-level scope at /builds/JuliaGPU/CuArrays.jl/test/iterator.jl:2
```

maleadt:
bors d+

jrevels:
Ah, thanks should be fixed now: 

bors try

maleadt:
Added some docs in https://github.com/JuliaGPU/CUDA.jl/commit/97cccad958cc3eefbdb627e8aadf0c1ee86381d5

PR #494, opened 2019-11-14
WIP: prepare cuArray for latest cuTENSOR version

springer13:
This PR is work-in-progress.

@maleadt @kshyatt @Jutho  could you please have a look at the tensor/wrapper.jl (especially the contraction function)?

I'm fully aware that this is not yet working, however, I think that I've wrapped the latest API properly and also updated the contraction function to reflect the new API.

The one thing that stands out most w.r.t. the latest cuTENSOR version is that all types no longer are pointers but structs of a fixed size. This requires that all those arguments must be passed in as pointers now.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,7 +1,7 @@
 include:
   - 'https://raw.githubusercontent.com/JuliaGPU/gitlab-ci/master/templates/v6.yml'
 
-image: juliagpu/cuda:10.1-cudnn7-cutensor-devel-ubuntu18.04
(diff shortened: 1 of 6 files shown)


kshyatt:
@maleadt is the CI set up for 0.3.0 yet? We could just run tests and see what happens.

maleadt:
I pushed a new CI image and enabled that here.
bors try

maleadt:
Pushed at https://github.com/JuliaGPU/CuArrays.jl/pull/502 (that one has easier CI).

Issue #498, opened 2019-11-15
CuArrays v1.4.6 fails to load, but v1.4.4 works fine


oschulz:
The recent update of CUDAdrv, CUDAapi, CUDAnative and CuArrays broke something, at least with my system configuration (I never had trouble with this system before, though).

I'm reporting this under CuArrays, since the symptom is that I can load CUDAdrv, CUDAapi and CUDAnative, but not CuArrays - the root cause may be in one of the other packages of course. I tried to debug, but didn't get anywhere. I hope the following information is helpful - I'd be glad for advice on how I can produce additional useful information to get to the bottom of this.

Using `CUDAdrv@4.0.3 CUDAapi@1.2.0 CUDAnative@2.5.4 CuArrays@1.4.4`, all is fine:

```
(v1.2) pkg> add CUDAdrv@4.0.3 CUDAapi@1.2.0 CUDAnative@2.5.4 CuArrays@1.4.4

julia> ENV["JULIA_DEBUG"] = "CUDAnative"; ENV["JULIA_CUDA_VERBOSE"] = "true";

julia> using CUDAdrv, CUDAapi, CUDAnative
[ Info: Precompiling CUDAdrv [c5f51814-7f29-56b8-a69c-e4d8f6be1fde]
[ Info: Precompiling CUDAapi [3895d2a7-ec45-59b8-82bb-cfc6a382f9b3]
[ Info: Precompiling CUDAnative [be33ccc6-a3ff-5ff2-a52e-74243cff1e17]
┌ Debug: Using LLVM v6.0.1
└ @ CUDAnative /tmp/user/juliadir/packages/CUDAnative/3Jwj2/src/CUDAnative.jl:143
┌ Debug: LLVM supports devices 2.0, 2.1, 3.0, 3.2, 3.5, 3.7, 5.0, 5.2, 5.3, 6.0, 6.1, 6.2 and 7.0; PTX 3.2, 4.0, 4.1, 4.2, 4.3, 5.0 and 6.0
└ @ CUDAnative /tmp/user/juliadir/packages/CUDAnative/3Jwj2/src/CUDAnative.jl:164
┌ Debug: Using CUDA driver v10.1.0 and toolkit v10.1.243
└ @ CUDAnative /tmp/user/juliadir/packages/CUDAnative/3Jwj2/src/CUDAnative.jl:169
[...]

PR #501, opened 2019-11-15
Don't needlessly populate the DL_LOAD_PATH.

maleadt:
Fixes https://github.com/JuliaGPU/CuArrays.jl/issues/498

```
julia> using CUDAdrv

julia> using Libdl

julia> Libdl.dlpath(Libdl.dlopen(:libcuda))
"/.singularity.d/libs/libcuda.so"

julia> ENV["JULIA_CUDA_VERBOSE"] = true
true

julia> using CuArrays
[ Info: Precompiling CuArrays [3a865a2d-5b23-5a0f-bc46-62713ec82fae]

julia> Libdl.dlpath(Libdl.dlopen(:libcuda))
"/.singularity.d/libs/libcuda.so"

julia> CuArray([1])
1-element CuArray{Int64,1,Nothing}:
 1

```

@oschulz I suggest you get rid of the multiple `libcuda.so`'s though :-)

### DIFF
--- a/src/CuArrays.jl
+++ b/src/CuArrays.jl
@@ -78,10 +78,16 @@ function __init__()
             lib = Symbol("lib$name")
             path = find_cuda_library(name, toolkit)
             libraries[name] = path
+
+            # only push the load path if we couldn't find the library
             if path !== nothing
-                dir = dirname(path)
-                if !(dir in Libdl.DL_LOAD_PATH)
-                    push!(Libdl.DL_LOAD_PATH, dir)
+                file = basename(path)
+                handle = first(split(file,'.'))
+                if Libdl.dlopen_e(handle) == C_NULL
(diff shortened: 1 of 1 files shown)


oschulz:
> @oschulz I suggest you get rid of the multiple libcuda.so's though :-)

It's not there by accident (for future reference, see #498 for the reason). :-)

oschulz:
Thanks a lot for the super-quick fix, @maleadt !

oschulz:
Could there be a v1.4.7 with this? I know, my setup is kinda unusual, but if possible, I'd keep it like that for now - the container has quite a few users (though not all of them use CUDA-features) and I don't want to break anything right now. We have a collaboration meeting in two weeks, and I want to show off Julia+CUDA running withing the container, if possible. :-)

PR #502, opened 2019-11-18
Updates for CUTENSOR 1.0

maleadt:


maleadt:
@kshyatt Bumped the test bounds to the correct values, can you run the tests now?

The tests now seem to trigger a bunch of CUTENSOR_UNSUPPORTED errors.

kshyatt:
bors try

kshyatt:
Hmmm all of a sudden we're not testing CUTENSOR!

maleadt:
> Hmmm all of a sudden we're not testing CUTENSOR!

we are, check the 1.2 builder. not all CI systems have a recent enough GPU, so it only forces to run them on one of them (1.2 since that's the current stable version)

kshyatt:
Ah cool, I looked at the wrong builder - thanks!

maleadt:
:tada:

Jutho:
Do I see correctly that the contraction tests are only ran for real `eltype`?

maleadt:
I ran into a bunch of "operation unsupported" errors, after which @kshyatt had a call with Paul and removed a bunch of tests, so I'm assuming that was the intention...

PR #512, opened 2019-11-26
Don't always time allocations.

maleadt:


maleadt:
@KristofferC FWIW, inferred code still looks OK here (I only checked the crucial methods of the BinnedPool, but that's where the heavy reuse of `buf` would likely introduce boxes).

baggepinnen:
This PR vastly improved the situation for me, thanks! I still occasionally get an
```
error in running finalizer: ErrorException("task switch not allowed from inside gc finalizer")
```
but training continues without problems

PR #515, opened 2019-11-26
RNN Failure hackaround

MikeInnes:
See https://github.com/FluxML/Flux.jl/issues/923. Right now, just testing this to confirm if it fixes the original issue. A nasty fix but hopefully this shows up infrequently enough that it's OK.

TODO unsafe_free the unused workspaces.

### DIFF
--- a/src/dnn/rnn.jl
+++ b/src/dnn/rnn.jl
@@ -11,7 +11,14 @@
 using LinearAlgebra: copy_transpose!
 
 function makeworkspace(size)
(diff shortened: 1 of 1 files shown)


maleadt:
You are still allocating the reserve after that, which might change the heuristic again. I have a cleaner prototype hackaround, let me push that here in a minute.

MikeInnes:
Ok, closing this for now so we can see your version.

PR #521, opened 2019-11-28
Adding missing `cudnnSetConvolutionGroupCount` to pointers.json

arhik:
This change is necessary for setting `groupcount` incase of depthwise and groupwise convolutions.

The original cudnn api is as follows:

`cudnnStatus_t cudnnSetConvolutionGroupCount(  `
`  cudnnConvolutionDescriptor_t    convDesc,  ` 
` int                             groupCount)`

### DIFF
--- a/res/wrap/pointers.json
+++ b/res/wrap/pointers.json
@@ -1695,6 +1695,10 @@
         "convDesc": null,
         "groupCount": "Ptr"
     },
+    "cudnnSetConvolutionGroupCount": {
+        "convDesc": null,
+        "groupCount": null
+    },
     "cublasCreate_v2": {
         "handle": "Ptr"
     },


maleadt:
Thanks! You'll need to change the wrappers themselves too, though. The `pointers.json` is a database used if you manually run `wrap.jl`, but we don't do that automatically and commit the resulting `libcudnn.jl`.

maleadt:
We have the API already: https://github.com/JuliaGPU/CuArrays.jl/blob/b64c6c0866ef978552d684c022132474ce2f7693/src/dnn/libcudnn.jl#L388-L392

arhik:
Sorry. Function pointer was returning null so I assumed its missing. I didn't realized its declared at other place. Thanks. 

> Why do you need this entry? We only need it to resolve pointer arguments, which this call doesn't have.

I was just following other definitions. I was curious to know the purpose of `null` declarations as a followup. You have shed some light on it. Interesting!

maleadt:
Yeah, this stuff isn't documented at all, purely internal for generating wrappers. `null` indicates this argument isn't special, and can be used as-is. In other cases, when dealing with pointers, the argument indicates what kind of pointer we're dealing with (a pointer to CPU memory, to GPU memory, or to potentially both).

PR #533, opened 2019-12-07
reenable complex tensor contraction tests

Jutho:
I've reenabled the complex tensor contraction tests, they work for most cases (even some that did not work before), except for the case where the conjugation flag is used (for either A and or B) in the specific cases `(NoA, NoB, Nc) in ((1,1,3), (1,2,3), (3,1,2))`. Not sure why this is. I've heard from @maleadt that @kshyatt had a talk with Paul about this?


maleadt:
CI failure was a timeout, so this is good to go. Thanks!

PR #541, opened 2019-12-16
Overhaul documentation and README.

maleadt:


maleadt:
FWIW, this is removing a bit of content from the README, but I'd rather forward people as quickly as possible to the shared documentation rather than maintain multiple READMEs.

maleadt:
` ERROR: LoadError: IOError: could not spawn `git rev-parse HEAD`: no such file or directory (ENOENT)`

PR #554, opened 2020-01-06
Improve error display.

maleadt:
Implement Base.showerror instead of show, use uniform output scheme.


maleadt:
ARM segfault unrelated.

PR #559, opened 2020-01-07
Use nightly Julia on ARM.

maleadt:
See if the LLVM upgrade fixes the segfault.


maleadt:
Oh huh we don't have nightlies for ARM.

PR #564, opened 2020-01-17
Basic tests for CuTensor type

kshyatt:
Looked like these weren't tested yesterday in the codecov...

### DIFF
--- a/test/tensor.jl
+++ b/test/tensor.jl
@@ -11,6 +11,26 @@ else
 
 @test has_cutensor()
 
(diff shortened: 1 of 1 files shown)


maleadt:
Looking at the coverage report, two additional ones:
- `cutensorCreate` in wrappers.jl can go, that's from previous CUTENSOR
- `src/CuArrays.jl` calls `CUDNN.cuda_version` twice; should be `CUTENSOR.cuda_version` for one of them

PR #568, opened 2020-01-22
Exclude some large kernels from coverage.

maleadt:


maleadt:
cc @kshyatt since you care about coverage. This builds on https://github.com/JuliaCI/CoverageTools.jl/pull/35

Issue #542, opened 2019-12-16
Invalid IR for reductions along trivial dimension


PhilipVinc:
**Describe the bug**
Performing reduction operations along some dimensions, where the destination array has fewer dimensions than the source produces invalid IR. This works with Base arrays. 

example:
```julia
julia> using CuArrays
julia> src = CuArrays.rand(10,10)
julia> dst_1 = similar(src, 10)
julia> dst_2 = similar(src, 10, 1)

julia> sum!(dst_2, src) # works
10×1 CuArray{Float32,2,Nothing}:
...

julia> sum!(dst_1, src) # fails

julia> sum!(collect(dst_2), collect(src)) # works
julia> sum!(collect(dst_1), collect(src)) # works
```

Both those methods work with Base arrays.

**To Reproduce**
The Minimal Working Example (MWE) for this bug:
```julia
using CuArrays
src = CuArrays.rand(10,10)
dst_1 = similar(src, 10)

# I guess the errors stems from Base.mapreduce, so...
Base.mapreducedim!(identity, +, dst_1, src)
```

**Environment details (please complete this section)**
Details on Julia:
```
julia> versioninfo()
Julia Version 1.3.0
Commit 46ce4d7933 (2019-11-26 06:09 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: Intel(R) Xeon(R) Silver 4116 CPU @ 2.10GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-6.0.1 (ORCJIT, skylake)
Environment:
  JULIA_DEPOT_PATH = :/opt/julia/global_depot/
  JULIA_LOAD_PATH = :/opt/julia/global_depot/environments/globalenv/


Julia packages:
 - CuArrays.jl: 1.5.0
 - CUDAnative.jl: 2.6.0
```

PhilipVinc:
If you can point me in the right direction, I would not mind fixing this

vchuravy:
The `mapreducedim` lives AFAIK in https://github.com/JuliaGPU/GPUArrays.jl/blob/master/src/mapreduce.jl

You can also try to follow the dispatch chain with `@which`, but it should lead you there.

PhilipVinc:
I just realised I had forgotten to include the error, so here it goes.

```julia
julia> # I guess the errors stems from Base.mapreduce, so...
[...]

PR #575, opened 2020-01-29
More adaption to GPUArrays

maleadt:
Fixes https://github.com/JuliaGPU/CuArrays.jl/issues/542


maleadt:
Flux broken because of https://github.com/FluxML/Flux.jl/pull/1008

PR #580, opened 2020-01-31
Upstream ldiv! overload

ChrisRackauckas:
DiffEqBase.jl has been carrying an ldiv! overload to make it work for awhile (https://github.com/JuliaDiffEq/DiffEqBase.jl/blob/master/src/init.jl#L148-L152), and I think it might be a good time to upstream it.


maleadt:
Thanks! Want to add a test?

ChrisRackauckas:
test added

maleadt:
@ChrisRackauckas The CI failure looks related?

xukai92:
Would it make sense to make https://github.com/JuliaGPU/CuArrays.jl/blob/master/src/linalg.jl take use of this?

xukai92:
Emmm. Together with `qr!`, this implementation seems to be slower than what we have in https://github.com/JuliaGPU/CuArrays.jl/blob/master/src/linalg.jl#L9-L13

ChrisRackauckas:
But that implementation cannot reuse factorizations, so this is a lot faster in practice.

xukai92:
Good point!

xukai92:
I guess we should keep both then. I use `\` in a case where no reuse could be taken.

ChrisRackauckas:
It's also possible there's a better way to do this, but I haven't found it 🤷‍♂

PR #598, opened 2020-02-18
Attempt at mixed precision for contraction

kshyatt:
DO NOT MERGE THIS

Paul (@springer13) agreed to take a look and see what's up with this sad PR that likely indicates I'm doing something wrong!

### DIFF
--- a/src/tensor/wrappers.jl
+++ b/src/tensor/wrappers.jl
@@ -207,9 +207,10 @@ function contraction!(
     alpha::Number, A::CuArray, Ainds::ModeType, opA::cutensorOperator_t,
                    B::CuArray, Binds::ModeType, opB::cutensorOperator_t,
     beta::Number,  C::CuArray, Cinds::ModeType, opC::cutensorOperator_t,
-                                                opOut::cutensorOperator_t,
+                                                opOut::cutensorOperator_t;
     pref::cutensorWorksizePreference_t=CUTENSOR_WORKSPACE_RECOMMENDED,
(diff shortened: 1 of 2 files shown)


src/tensor/wrappers.jl
```diff
@@ -251,8 +251,8 @@ function contraction!(
             cutensorInitContractionPlan(handle(), plan, desc, find, sizeof(workspace))
 
             cutensorContraction(handle(), plan,
-                                T[alpha], A, B,
-                                T[beta],  C, C,
+                                compute_type[alpha], A, B,
```
springer13: The behavior w.r.t. the scalars in cuTENSOR is, unfortunately, slightly more involved:
https://docs.nvidia.com/cuda/cutensor/user_guide.html#scalar-types

Thus, it's not a 1:1 mapping. Maybe this is causing an issue?

kshyatt: I thought that the `alpha` and `beta` constants are of type `typeCompute` in `contraction`? That's what the docs [here](https://docs.nvidia.com/cuda/cutensor/api/cutensor.html#cutensorcontraction) implied to me.

kshyatt: Ah, I see - with a change things seem to be doing better, although mysteriously I'm getting a `not supported` error for all `Float32` tensors with a compute type of `Float16`...

springer13: I agree, the documentation is somewhat confusing; we'll add a reference to https://docs.nvidia.com/cuda/cutensor/user_guide.html#scalar-types as part of the contraction docs.

Could you please run the contraction with CUTENSOR_DEBUG=1 ? Which GPU are you computing on; FP16 support only starts with Volta.

PR #604, opened 2020-02-24
Actually test error throwing for lsvqr

kshyatt:
Can't test mismatched dimensions if the GPU doesn't know about them

### DIFF
--- a/test/sparse_solver.jl
+++ b/test/sparse_solver.jl
@@ -47,9 +47,15 @@ k = 1
         A = sparse(rand(elty,n,n))
         b = rand(elty,m)
         x = zeros(elty,n)
+        d_A   = CuSparseMatrixCSR(A)
+        d_b   = CuArray(b)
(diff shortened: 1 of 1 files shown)


maleadt:
`CUBLAS_STATUS_ALLOC_FAILED` -- these calls allocate? the calls should be wrapped in an `@allocates` like some CURAND calls are.

kshyatt:
Oh, yeah, because they are making copies. Sorry about that.

maleadt:
Basically this but with a different code: https://github.com/JuliaGPU/CuArrays.jl/blob/0cdf32ce11263446c17297f62157e5f944912e21/src/rand/wrappers.jl#L13-L19

maleadt:
Also, if you know how much memory the call would probably allocate, pass that as the `nb` kwarg. Otherwise we might end up deallocating all cached memory.

PR #605, opened 2020-02-24
RFC: provide a function to generate a contraction plan, and let contraction use it

kshyatt:
The idea is to let people take advantage of CUTENSOR's plan generation and reuse, if they want. If they don't provide a plan, CUTENSOR will just make one for them.

### DIFF
--- a/src/tensor/wrappers.jl
+++ b/src/tensor/wrappers.jl
@@ -207,9 +207,10 @@ function contraction!(
     alpha::Number, A::CuArray, Ainds::ModeType, opA::cutensorOperator_t,
                    B::CuArray, Binds::ModeType, opB::cutensorOperator_t,
     beta::Number,  C::CuArray, Cinds::ModeType, opC::cutensorOperator_t,
-                                                opOut::cutensorOperator_t,
+                                                opOut::cutensorOperator_t;
     pref::cutensorWorksizePreference_t=CUTENSOR_WORKSPACE_RECOMMENDED,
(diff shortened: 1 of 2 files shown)


test/runtests.jl
```diff
@@ -38,7 +38,7 @@ device!(pick.dev)
 CuArrays.allowscalar(false)
 
 CuArrays.enable_timings()
-
+#=
```
maleadt: I think we need to support arguments to `runtests.jl` a la base for our sanity :-)

kshyatt: oh my god derp

kshyatt: ban me

kshyatt:
OK, tests look happier now. Do you have thoughts on the design here? I know I need to write a test for this use case.

kshyatt:
OK, we have tests and all that. Thoughts?

src/tensor/wrappers.jl
```diff
@@ -239,25 +240,73 @@ function contraction!(
...
+            end
+            cutensorContraction(handle(), plan_ref,
+                                T[convert(T, alpha)], A, B,
+                                T[convert(T, beta)],  C, C,
                                 workspace, sizeof(workspace), stream)
         end
     return C
 end
 
+function contraction_plan(
```
maleadt: maybe `plan_contraction` re. FFTW?

kshyatt:
Well, the test seems to be passing. I'll change the name as you suggested.

kshyatt:
Tests passed, going to merge.

PR #610, opened 2020-02-25
Include semi-documented contraction algo options

kshyatt:
In fact, there are more algorithms than we include now:

https://docs.nvidia.com/cuda/cutensor/api/types.html#cutensoralgo-t

`values >= 0 correspond to certain sub-algorithms of GETT`

I checked and currently the maximum number of algorithms function returns 9. Maybe there's a runtime way to set this up?


maleadt:
If those algorithms aren't listed in the header, are we "supposed" to use them (stability, correctness, and whatnot)? Modifying the generated files doesn't seem ideal either.

kshyatt:
@springer13 is there some way we could get these doc-ed in CUTENSOR's documentation/listed in the header for the auto-generation script?

PR #612, opened 2020-02-29
Specialized _var and _std functions

### LINKED ISSUE #411: Statistics functions (var,std) slow
var/std seem to fall back on indexing the CuArray, instead of a vectorized implementation:
```
julia> var(c,dims=1)
┌ Warning: Performing scalar operations on GPU arrays: This is very slow, consider disallowing these operations with `allowscalar(false)`
└ @ GPUArrays ~/.julia/packages/GPUArrays/J4c3Q/src/indexing.jl:16
1×4 CuArray{Float32,2}:
 0.0646661  0.167767  0.0156315  0.14948
```
This makes it slower than cpu:
```
julia> c1 = rand(1000,1000);
julia> c2 = CuArray(c1);
julia> @time std(c1,dims=1);
  0.001371 seconds (12 allocations: 16.250 KiB)
julia> @time std(c2,dims=1);
  7.945594 seconds (3.01 M allocations: 138.023 MiB, 0.39% gc time)
julia> @time CuArray(std(Array(c2),dims=1))
  0.003153 seconds (23 allocations: 7.661 MiB)
```

merckxiaan:
I tried writing a CuArrays version for _var and _std which fixes #411.
@kshyatt [mentioned the issue actually lies in centralize_sumabs2!](https://github.com/JuliaGPU/CuArrays.jl/issues/411#issuecomment-539549996) but I'm not quite sure what that function does or how it should be done on gpu.
Perhaps this solution suffices?


maleadt:
Thanks! These definitions seems OK, if they execute without scalar iteration of course. Could you add some tests comparing against results with Base.Array?

merckxiaan:
I added some, please let me know if these need changes or need to be expanded.

test/statistics.jl
```diff
@@ -0,0 +1,16 @@
+@testset "Statistics" begin
+
+using CuArrays
```
maleadt: No need for this import.

test/statistics.jl
```diff
@@ -0,0 +1,16 @@
...
+using Statistics
+
+@testset "std" begin
+    A = rand(10, 1, 2)
+    cuA = cu(A)
+
+    @test std(cuA[:, 1, 1]) ≈ std(A[:, 1, 1])
```
maleadt: runtests.jl defines a compare function that automatically constructs a CuArray, performs the method, and compares back on the CPU. so you just need `compare(std, rand(10, 1, 2))`.

test/statistics.jl
```diff
@@ -0,0 +1,16 @@
...
+    @test std(cuA) ≈ std(A)
+    @test std(cuA, corrected=true) ≈ std(cuA, corrected=true, dims=:) ≈ std(A, corrected=true)
+    @test collect(std(cuA, dims=1)) ≈ std(A, dims=1)
+end
```
maleadt: Maybe add a test for `var` too? I know one calls the other, but just to be explicit.

test/statistics.jl
merckxiaan: I found CuArrays uses `testf` which, in turn, depends on `compare`. This is probably what you want?

merckxiaan:
I made the changes and also realized the tests hadn't actually been added to the runtests.jl so that's fixed as well 😅.
Let me know if it needs further changes.
Thanks for the guidance!

maleadt:
bors try

maleadt:
Great, let's merge this. Thanks for the PR!

PR #619, opened 2020-03-05
NNlib batched_mul!

mcabbott:
This hooks up https://github.com/FluxML/NNlib.jl/pull/100 for CuArrays. 

I've added an extremely simple test, but could find the right place to insert this into test/blas.jl perhaps if that would be better?

Tests fail locally, before this PR, but I'm not sure why.

I think that `_GemmFloat == CUBLAS.CublasFloat`, except that this isn't loaded yet. Is there a neater way?


maleadt:
> I've added an extremely simple test, but could find the right place to insert this into test/blas.jl perhaps if that would be better?

It's an nnlib specific interface, so that place seems fine.



> Tests fail locally, before this PR, but I'm not sure why.

Please open an issue for that then. Let's see what CI does here:
bors try



> I think that `_GemmFloat == CUBLAS.CublasFloat`, except that this isn't loaded yet. Is there a neater way?

Move the include of `nnlib.jl` until after the submodules (including CUBLAS)?

mcabbott:
Thanks for having a look, I've moved the `include("nnlib.jl")` as suggested. 

About tests, is there a list of what needs to be on master? I failed again, but can try more seriously once back in front of my desktop, probably in 2 weeks.

maleadt:
> About tests, is there a list of what needs to be on master? I failed again, but can try more seriously once back in front of my desktop, probably in 2 weeks.

Generally CuArrays, CUDAnative, CUDAdrv and GPUArrays. CuArrays also ships a Manifest with a known-good combination, so you can just clone CuArrays and run with `julia --project` in that dir.

PR #623, opened 2020-03-10
Use a library handle for each thread/device.

maleadt:
https://github.com/cupy/cupy/issues/1109


maleadt:
@vchuravy @kshyatt This seems like the correct way to go, but it would complicate, say, creating a contraction plan and using that in a threaded region. But we kind-of have that problem already since apparently tasks can migrate threads?

maleadt:
Going to merge this because the locks are required (and I want to introduce some threaded tests), but feel free to comment or review after that.

PR #627, opened 2020-03-11
Thread safety of memory allocator

maleadt:


maleadt:
Ah, finally. CI failures are some CUBLAS allocations that aren't properly guarded by `@extalloc`.

Issue #629, opened 2020-03-12
mul! failing for adjoint sparse matrices 


maximilian-gelbrecht:
MWE 
```Julia
using CuArrays
using SparseArrays
using LinearAlgebra

const n = 100

M = CuArray(diagm(0=>ones(n), 1=>-1*ones(n-1), -1=>-1*ones(n-1)))
M2 = sparse(M);
V = CuArray(rand(100));
adjoint(M2)*V
```
yields 

```
UndefVarError: transA not defined

Stacktrace:
 [1] mul!(::CuArray{Float64,1,Nothing}, ::Adjoint{Float64,CuArrays.CUSPARSE.CuSparseMatrixCSR{Float64}}, ::CuArray{Float64,1,Nothing}) at /home/USERNAME/.julia/packages/CuArrays/A6GUx/src/sparse/interfaces.jl:14
 [2] *(::Adjoint{Float64,CuArrays.CUSPARSE.CuSparseMatrixCSR{Float64}}, ::CuArray{Float64,1,Nothing}) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.3/LinearAlgebra/src/matmul.jl:113
 [3] top-level scope at In[3]:1
```

just from inspecting the code, this seems to me like an obvious error in the code, or isn't it?

it is:
```Julia
LinearAlgebra.mul!(C::CuVector{T},adjA::Adjoint{<:Any,<:CuSparseMatrix},B::CuVector) where {T} = mv!('C',one(T),parent(transA),B,zero(T),C,'O')
```
but it should be 
```Julia
LinearAlgebra.mul!(C::CuVector{T},adjA::Adjoint{<:Any,<:CuSparseMatrix},B::CuVector) where {T} = mv!('C',one(T),parent(adjA),B,zero(T),C,'O')
``` 
or is there more to it? It did not try it out yet.

maleadt:
Seems like a trivial issue indeed, and we should also add a test for that code.

PR #630, opened 2020-03-12
fix sparse adjoint mul

maximilian-gelbrecht:
as reported in https://github.com/JuliaGPU/CuArrays.jl/issues/629   

one could still add an additional test.


CarloLucibello:
> one could still add an additional test.

definitely!

PR #631, opened 2020-03-13
Trap and retry CUBLAS allocation failures.

maleadt:
This retries CUBLAS API calls when they fail to allocate, as indicated by a CUBLAS_STATUS_ALLOC_FAILED. I just wrapped those methods that I saw fail on CI:
- https://gitlab.com/JuliaGPU/CuArrays.jl/-/jobs/471002172
- https://gitlab.com/JuliaGPU/CuArrays.jl/-/jobs/471049468
- https://gitlab.com/JuliaGPU/CuArrays.jl/-/jobs/471049476

There's bound to be others, though, but even for these calls the NVIDIA docs don't mention ALLOC_FAILED as a valid return code (indicating to me they shouldn't allocate?). I don't know, as I'm not familiar with this specific library.

@kshyatt could you have a look, since you wrapped cublasXt and presumably are using it? This is also what I meant in https://github.com/JuliaGPU/CuArrays.jl/pull/604, by the way.

### DIFF
--- a/src/blas/wrappers.jl
+++ b/src/blas/wrappers.jl
@@ -43,6 +43,15 @@ version() = VersionNumber(cublasGetProperty(CUDAapi.MAJOR_VERSION),
                           cublasGetProperty(CUDAapi.MINOR_VERSION),
                           cublasGetProperty(CUDAapi.PATCH_LEVEL))
 
+macro allocates(ex)
+  quote
+    CuArrays.extalloc(check=err->isa(err, CUBLASError) &&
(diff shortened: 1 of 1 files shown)


maleadt:
Two more from https://gitlab.com/JuliaGPU/CuArrays.jl/-/jobs/471216418

maleadt:
I was thinking we should probably do this for all API calls -- should be cheap enough if we don't try/catch. That's assuming state doesn't get corrupted if an allocation fails.

maleadt:
Superseded by https://github.com/JuliaGPU/CuArrays.jl/pull/633

PR #633, opened 2020-03-14
Use a return value-based retry scheme for all APIs failing to allocate.

maleadt:


maleadt:
I wonder if we could use the same for API initialization; when the API throws e.g. CUBLAS_STATUS_NOT_INITIALIZED.

PR #634, opened 2020-03-15
Provide 5-arg mul!

haampie:
Since Julia 1.3 we have a 5-arg mul! for BLAS, but somehow it has not made its way into CuArrays.

In principle this should not be a breaking change, however, I've commented out `LinearAlgebra.lmul!` because it is not provided by Julia base as far as I know. So that is breaking, but I don't think anybody uses it (?).

Finally it fixes an oversight in `LinearAlgebra.mul!(C::CuMatrix{T}, adjA::Adjoint{<:Any, <:CuMatrix{T}}, adjB::Adjoint{<:Any, CuMatrix{T}})` where no `<:` was used 😅

### DIFF
--- a/src/blas/linalg.jl
+++ b/src/blas/linalg.jl
@@ -87,14 +87,16 @@ LinearAlgebra.mul!(Y::CuVector{T}, A::Adjoint{<:Any, <:CuMatrix{T}}, B::CuVector
     gemv_wrapper!(Y, 'C', A.parent, B, promote_alpha_beta(a, b, T)...)
 
 # Fix Julia 1.3.0 ambiguities... they're fixed in 1.3.1 thanks to https://github.com/JuliaLang/julia/pull/33743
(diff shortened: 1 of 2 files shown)


haampie:
Hm, there's ambigious methods when taking `a` and `b` as `::Number`. Simple fix is to replace that type by the eltype `T`, but that's not such a nice api.

maleadt:
Thanks! Next release is going to be breaking so go ahead and remove those methods.

haampie:
Should there be a deprecation warning for `lmul!`?

maleadt:
> Should there be a deprecation warning for `lmul!`?

Seems weird to provide a deprecation for a method in another module? We're just implementing the interface here; if upstream has deprecated/removed something, shouldn't we just follow suit?

haampie:
Ok, so I've worked my way through tons of ambiguity warnings and added tests for all combinations of eltypes, transposes and adjoints for `mul!`. Ready for review!

maleadt:
bors try

maleadt:
LGTM, coverage is perfect too. Would it make sense to guard the definitions for 1.3.0 in a `if VERSION` block, to make sure we can safely remove them once we bump the Julia requirement?

bord d+

haampie:
Ok, done :)

PR #639, opened 2020-03-18
Let cublasXT use host memory

kshyatt:
the whole point of cublasXT is to allocate host arrays that are too big for one GPU, and then it farms out the chunks for you. not sure I did the wrapper updates right. seeing a lot of ugly test failures if anyone else wants to take a run at this.

### DIFF
--- a/res/wrap/pointers.json
+++ b/res/wrap/pointers.json
@@ -4460,9 +4460,9 @@
         "m": null,
         "n": null,
         "alpha": "PtrOrCuPtr",
-        "A": "CuPtr",
+        "A": "PtrOrCuPtr",
         "lda": null,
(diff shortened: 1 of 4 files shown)


maleadt:
Only issue was an OOM, so that actually looks rather good? Or don't we have good test coverage for cublasXt?

kshyatt:
The difference is that our previous coverage was with GPU pointers only. I
had to modify the wrappers to accept CPU pointers too (that's actually the
normal operation mode of cublasXt) and then add tests for it.

kshyatt:
Tests passed for the `pool` tests, there was just a job timeout. Good to merge?

maleadt:
I restarted the rests.
bors r+

PR #640, opened 2020-03-18
Test Julia 1.4.

maleadt:
@haampie Looks the 5-arg mul! fixes broke 1.4. Part of that is that is due to bugs in GPUArrays' generic_matmul kernel, e.g. when calling `CuArray(transpose(rand(Float32, 20))) * CuArray(rand(Float32, 20))`, but CuArrays should be dispatching to CUBLAS in that case. I guess the `mul!` methods in GPUArrays need updating too?

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -25,6 +25,13 @@ julia:1.3:
   variables:
     CI_THOROUGH: 'true'
 
+julia:1.4:
+  extends:
+    - .julia:1.4
+    - .test
+  tags:
+    - nvidia
+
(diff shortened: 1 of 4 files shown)


maleadt:
Uhh, how did tests pass here? The 1.4 logs are full of errors.

maleadt:
Okay, this restores CI. I reverted your pr, @haampie, as it also seemed to fail on 1.4 (despite the PR you listed, and the changes you mentioned, are definitely part of 1.3.1). We can revisit that after the master branch passes tests again.

PR #641, opened 2020-03-18
5-arg mul! with 3-arg mul! support

haampie:
Turns out I was not right about that static if for julia 1.3.0 (the relevant commit in base was merged before the release of 1.3.1, but only made its way into julia from 1.4.0-rc1).

Otherwise this should work after https://github.com/JuliaGPU/GPUArrays.jl/pull/253 is merged, because GPUArrays is stealing 3-arg mul! from CuArrays at the moment.

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -34,9 +34,9 @@ version = "4.0.0"
 
 [[CUDAdrv]]
 deps = ["CEnum", "CUDAapi", "Printf"]
-git-tree-sha1 = "f176b994a6e4c70aafa626cbf825aa9c34adc9e6"
+git-tree-sha1 = "9db0ff78ac601ca66c85f3bededdbacc3d1c9bdb"
 uuid = "c5f51814-7f29-56b8-a69c-e4d8f6be1fde"
-version = "6.1.0"
+version = "6.2.0"
 
 [[CUDAnative]]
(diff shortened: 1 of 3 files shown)


maleadt:
Once we merge the GPUArrays  one, do an `add GPUArrays#master` here so that we can run CI.

haampie:
I believe we need the master version of CUDAnative as well because of CUDAapi 4 compat.

Could you bors try this @maleadt?

maleadt:
> I believe we need the master version of CUDAnative as well because of CUDAapi 4 compat.

The Manifest already listed a specific git sha. I prefer not to use just `master` because that might hamper reproducibility.

haampie:
Ok, fixed that

maleadt:
bors try

maleadt:
bors d+

haampie:
bors try

haampie:
bors try-

maleadt:
bors try

haampie:
Oof... there's a Julia 1.4.0-x specific inconsistency / ambiguity for one last single mul! signature... Will work around it here and then make a PR for base

haampie:
bors try

maleadt:
Light at the end of the tunnel!
bors r+

haampie:
:relieved:

PR #656, opened 2020-03-28
Add support for uniform random UInt32

### LINKED ISSUE #655: rand(UInt32) gives low-quality results
**Describe the bug**
rand(UInt32) gives low-quality results

**To Reproduce**
The Minimal Working Example (MWE) for this bug:
```julia
sum(isodd.(CuArrays.rand(UInt32, 10000)))
```

**Expected behavior**
Values should be evenly distributed. Instead there are--for example--very few odd numbers

chrstphrbrns:
fix #655

### DIFF
--- a/src/rand/random.jl
+++ b/src/rand/random.jl
@@ -57,8 +57,12 @@ end
 ## in-place
 
 # uniform
-const UniformType = Union{Type{Float32},Type{Float64}}
-const UniformArray = CuArray{<:Union{Float32,Float64}}
+const UniformType = Union{Type{Float32},Type{Float64},Type{UInt32}}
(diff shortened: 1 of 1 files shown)


maleadt:
Thanks! The docs don't mention uniform distribution, but I guess that's the case for this call?
bors try

chrstphrbrns:
> Thanks! The docs don't mention uniform distribution, but I guess that's the case for this call?
> bors try

Yep

`The curandGenerate() function is used to generate pseudo- or quasirandom bits of output for XORWOW, MRG32k3a, MTGP32, MT19937, Philox_4x32_10 and SOBOL32 generators. Each output element is a 32-bit unsigned int where all bits are random. `

PR #675, opened 2020-04-06
Allow scalar indexing where necessary and add a few tests

kshyatt:


test/solver.jl
```diff
@@ -320,6 +321,11 @@ k = 1
         @test size(d_F) == size(A)
         @test size(d_F.Q, 1) == size(A, 1)
         @test det(d_F.Q) ≈ det(collect(d_F.Q * CuMatrix{elty}(I, size(d_F.Q)))) atol=tol*norm(A)
+        @test d_F.Q[1, 1] ≈ qra.Q[1, 1]
```
maleadt, line 324: Why should this succeed without requiring the user to `@allowscalar` itself? I.e., shouldn't the getindex method do an `assertscalar` instead of `@allowscalar`?

kshyatt: I was following the existing style in the method, I can change it if you like

maleadt: Assuming this getindex is indicative of bad scalar code, I think it should error.

test/solver.jl
```diff
@@ -321,9 +321,13 @@ k = 1
         @test size(d_F) == size(A)
         @test size(d_F.Q, 1) == size(A, 1)
         @test det(d_F.Q) ≈ det(collect(d_F.Q * CuMatrix{elty}(I, size(d_F.Q)))) atol=tol*norm(A)
-        @test d_F.Q[1, 1] ≈ qra.Q[1, 1]
+        # @allowscalar didn't work... and scalar indexing needed for sprint
```
maleadt, line 324: You should be able to do `CuArrays.@allowscalar begin ... end`, we use it for a bunch of other tests.

PR #683, opened 2020-04-15
Enable using host memory in CUTENSOR

kshyatt:
This *does* work, as long as you register the host memory and make sure to sync after the call to CUTENSOR (without the sync, sometimes the result isn't populated in time).

I can change the input types from `AbstractArray` if you feel that's too loose.


kshyatt:
This guy needs https://github.com/JuliaGPU/CUDAdrv.jl/pull/230 btw whoops

maleadt:
> This guy needs [JuliaGPU/CUDAdrv.jl#230](https://github.com/JuliaGPU/CUDAdrv.jl/pull/230) btw whoops

Add it to the manifest then(`] add CUDAdrv#GIT_HASH` -- or does it record the git hash when adding a branch?)

kshyatt:
I was going to wait to add it to the Manifest until after the squash and merge I presume is coming for that other branch :). Maybe the Manifest should update automatically but I'd probably manage to commit a non-existent commit.

maleadt:
> Maybe the Manifest should update automatically but I'd probably manage to commit a non-existent commit.

It does, but only for releases. Waiting for the merge is probably better though :slightly_smiling_face:

PR #689, opened 2020-04-22
skip @cuda call when total_threads is zero

AStupidBear:
This can fix concatenating of empty `CuArray`s. #317


maleadt:
It should fail earlier already: https://github.com/JuliaGPU/GPUArrays.jl/blob/ff47dcc17c136a55b82dbf70822a45c27d1446e3/src/device/execution.jl#L59-L66

I'm not entirely sure we want this though. On the other hand no real reasons not to... Thoughts, @vchuravy?

PR #690, opened 2020-04-22
fix 1d convolution

AStupidBear:
Since CUDNN does not support 1D convolution, `Conv` in `Flux` will give a CUDNNError if the `size` is one dimensional.
```
julia> c = Conv((2, ), 3 => 4) |> gpu
Conv((2,), 3=>4)

julia> c(gpu(rand(10, 3, 2)))
ERROR: CUDNNError(code 9, CUDNN_STATUS_NOT_SUPPORTED)
```

We have to reshape the `CuArray` / `PoolDims` / `DenseConvDims` to 4D before feeding to CUDNN.

### DIFF
--- a/src/dnn/nnlib.jl
+++ b/src/dnn/nnlib.jl
@@ -41,13 +41,23 @@ end
 
 # Convolution
 
+fix1d(x) = x
+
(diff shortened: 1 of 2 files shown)


maleadt:
Could you elaborate, and add that as a comment to the `fix1d` function?

maleadt:
CI failure unrelated.

PR #695, opened 2020-04-27
adds wrappers for syevjBatched/heevjBatched family of CUSOLVER functions

electronsandstuff:
Implemented batch hermitian/symmetric eigensolving from cuSolver.  Used same solver default parameters as sygvj which converges to machine precision for the 15x15 test matrices in the unit tests.

### DIFF
--- a/src/solver/wrappers.jl
+++ b/src/solver/wrappers.jl
@@ -823,3 +823,61 @@ for (jname, bname, fname, elty, relty) in ((:sygvj!, :cusolverDnSsygvj_bufferSiz
         end
     end
 end
+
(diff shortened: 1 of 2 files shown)


maleadt:
Thanks! Could you add some tests?

maleadt:
Thanks!
bors try

maleadt:
bors try

electronsandstuff:
It looks like it got stuck on a different test than last time.  It was Julia:1.4 now it's Arm64.

maleadt:
The ARM can happen, it shouldn't cause bors to timeout. I wonder if our CI times have increased lately, it sure feels like it. Let's check another time to be sure:
bors r+

maleadt:
Not sure what's up with CI. I rebooted the servers.
bors try

maleadt:
CI is flaky, but that doesn't look to be related to this PR. Let's merge this.

PR #706, opened 2020-05-05
Add correct batch size for RNN hidden layer

### LINKED ISSUE #1114: RNN on GPU fails on first backward call
In the following minimal RNN example, first call to `Flux.train!` fails with `CUDNN_STATUS_BAD_PARAM`. Same error raised either cell is `GRU`, `RNN` or `LSTM`.

```julia
using Flux
using Statistics

rnn = Chain(GRU(16, 8),
  Dense(8,1, σ),
  x -> reshape(x,:))

X = [rand(16,10) for i in 1:20]
Y = rand(10,20) ./ 10

rnn = rnn |> gpu
X = gpu(X)
Y = gpu(Y)

θ = Flux.params(rnn)
loss(x,y) = mean((Flux.stack(rnn.(X),2) .- y) .^ 2f0)
opt = ADAM(1e-3)
size(rnn[1].state)
Flux.reset!(rnn)
size(rnn[1].state)
Flux.train!(loss, θ, [(X,Y)], opt)
size(rnn[1].state)
loss(X,Y)
```

It can be observed that both prior and after reset!, rnn state is of size (8), while after a call to train! on GPU, state becomes the expected proper size (8,10). After each call to `reset!`, the CUDNN_STATUS_BAD_PARAM error pops out after first call to train!, but subsequent ones are fine as the state size stays (8,10). Can't confirm whether that state size is the root cause, but appears closely tied to the bug. Also, a call to `loss(X,Y)` results in a proper state dimension of (8,10). Running on CPU doesn't result in any error/warning. 

Pkg info (same error also raised wih latest Zygote release): 

```
  [4f1ea46c] AWSCore v0.6.9
  [1c724243] AWSS3 v0.6.10
  [fbb218c0] BSON v0.2.5
  [336ed68f] CSV v0.6.1
  [3a865a2d] CuArrays v2.0.1
  [a93c6f00] DataFrames v0.20.2
  [587475ba] Flux v0.10.4
  [28b8d3ca] GR v0.48.0
  [91a5bcdd] Plots v1.0.5
  [2913bbd2] StatsBase v0.32.2
  [e88e6eb3] Zygote v0.4.15 #master (https://github.com/FluxML/Zygote.jl.git)
  [10745b16] Statistics
```

CUDA: 10.1.168
CUDNN: 7.6.5

DhairyaLGandhi:
Fixes https://github.com/FluxML/Flux.jl/issues/1114

The context here is that on the first call to the layer (also simulated via calling the `Flux.reset!` on the structure), the gradients for the hidden layer were returned as the same size as the state at the time (a vector), whereas on stepping through time, the size of the state would change. This needs to be accounted for while sending the correct size back. We do this automatically for subsequent calls, only with the first run, is when we see this issue.

@MikeInnes  @maleadt, would this be an alright fix?

### DIFF
--- a/src/dnn/rnn.jl
+++ b/src/dnn/rnn.jl
@@ -101,6 +101,7 @@ end
 hBatch(x::AbstractVector, h::CuVector) = h
 hBatch(x::AbstractMatrix, h::CuVector) = h .*CuArrays.ones(1, size(x, 2))
 hBatch(x::AbstractMatrix, h::CuMatrix) = h .*CuArrays.ones(1, size(h,2) == 1 ? size(x,2) : 1)
+hBatch(x, ::Nothing) = nothing
 
 function forward(rnn::RNNDesc{T}, x::CuArray{T}, h_::CuArray{T}, c_ = nothing, train = Val{false}) where T
   h = hBatch(x, h_)
@@ -197,7 +198,7 @@ function pullback(rnn::RNNDesc{T}, x::CuArray{T}, h::CuArray{T}, c::CuArray{T}) 
   return (y, ho, co), function (dy, dho, dco)
     h_ = CUDNN.hBatch(x, h)
     c_ = CUDNN.hBatch(x, c)
-    dx, dh, dc = CUDNN.backwardData(rnn, y, dy, dho, dco, h_, c_, reserve)
+    dx, dh, dc = CUDNN.backwardData(rnn, y, dy, CUDNN.hBatch(y,dho), dco, h_, c_, reserve)
     (dWi, dWh), db = CUDNN.backwardWeights(rnn, x, h_, y, reserve)
     return (x = dx, h = dh, c = dc, Wi = dWi, Wh = dWh, b = db)
   end
--- a/src/dnn/rnn.jl
+++ b/src/dnn/rnn.jl
@@ -101,7 +101,6 @@ end
 hBatch(x::AbstractVector, h::CuVector) = h
 hBatch(x::AbstractMatrix, h::CuVector) = h .*CuArrays.ones(1, size(x, 2))
 hBatch(x::AbstractMatrix, h::CuMatrix) = h .*CuArrays.ones(1, size(h,2) == 1 ? size(x,2) : 1)
-hBatch(x, ::Nothing) = nothing
 
 function forward(rnn::RNNDesc{T}, x::CuArray{T}, h_::CuArray{T}, c_ = nothing, train = Val{false}) where T
   h = hBatch(x, h_)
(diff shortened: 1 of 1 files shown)


maleadt:
Maybe add a test?

MikeInnes:
As I understand it, `ho` should already have been expanded (via `hbatch`) during the forwards pass. Its gradient `dho` should have the same size as `ho`, in which case AFAICT this patch shouldn't be needed.

I don't think this PR would do any harm, but that seems fishy and it may be worth understanding why it happens in case something else is going wrong.

dhairyagandhi96:
https://github.com/JuliaGPU/CuArrays.jl/blob/eac33f8f383f473c31bd879611fad15ee56671bb/src/dnn/rnn.jl#L188 and others is where that expansion happens. This is just to apply the same treatment as `h` to `dh`.

MikeInnes:
I believe that line mirrors [this one](https://github.com/JuliaGPU/CuArrays.jl/blob/eac33f8f383f473c31bd879611fad15ee56671bb/src/dnn/rnn.jl#L106) which does happen in the forward pass (we just do it again in the backwards pass because we don't have access to the expanded `h`).

It would be worth printing out the size of `ho` and `dho` at the start of the pullback. Pullbacks are supposed to be able to rely on them being the same size (and if they are then there's no issue).

dhairyagandhi96:
Found this
```Julia
(size(ho), size(dho)) = ((8, 10), (8,))
```
This was when the model was `reset!`d

All subsequent calls have the same size. I am assuming its because `h` is a vector that we get a vector back? here https://github.com/JuliaGPU/CuArrays.jl/blob/eac33f8f383f473c31bd879611fad15ee56671bb/src/dnn/rnn.jl#L185

dhairyagandhi96:
What should the test actually be? Just calling into the `backwardData` function?

MikeInnes:
If `h` is a vector then `dh` should also be a vector; I'm not sure if we handle that correctly. However, it seems likely to be independent of the current issue.

The fact that `size(dho) != size(ho)` implies that some other function is getting called with a matrix (`ho`) but producing a vector (`dho`) as a gradient, which is wrong. Ideally we'd fix that function, whatever it is.

jeremiedb:
@DhairyaLGandhi @MikeInnes Have you had a moment to look at the above? 

Using the following latest release, the issue seems not to be reproducible anymore: 
```julia
  [3a865a2d] CuArrays v2.2.2
  [a93c6f00] DataFrames v0.21.2
  [587475ba] Flux v0.10.4
  [28b8d3ca] GR v0.50.1
  [91a5bcdd] Plots v1.4.3
  [2913bbd2] StatsBase v0.33.0
  [e88e6eb3] Zygote v0.4.22
```

However, I couldn't identify what change to either Flux, CuArrays or Zygote may have solved the issue. Any pointer would be welcome.

DhairyaLGandhi:
Am I right in understanding that the mwe in FluxML/Flux.jl#1114 works as expected now?

jeremiedb:
Yes!

DhairyaLGandhi:
Wonder if the Adapt issue is related then, incorrect size in the backwards pass could result from an `ndims` of a `Transpose` or similar.

DhairyaLGandhi:
Ref JuliaGPU/Adapt.jl#24, I suppose

PR #727, opened 2020-05-23
Add inplace rand test for Cuint

kshyatt:
Looks [uncovered](https://codecov.io/gh/JuliaGPU/CuArrays.jl/src/master/src/rand/random.jl#L64)


maleadt:
Move to CUDA.jl? Sorry for creating some extra work here...

